PR #6, opened 2020-03-25
Fixes a bug when subfolder has spaces

fdiaz:
My current approach to traversing the file structure has a bug, it doesn't work when a subfolder has spaces.

This fixes that bug.

### DIFF
--- a/Sources/SwiftInspectorKit/FileManager+SwiftFiles.swift
+++ b/Sources/SwiftInspectorKit/FileManager+SwiftFiles.swift
@@ -40,7 +40,7 @@ extension FileManager {
 
     guard let enumerator = self.enumerator(
       at: baseURL,
-      includingPropertiesForKeys: [.isDirectoryKey],
+      includingPropertiesForKeys: nil,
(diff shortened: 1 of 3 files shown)


Sources/SwiftInspectorKit/FileManager+SwiftFiles.swift
```diff
@@ -38,10 +38,17 @@ extension FileManager {
 
     var swiftFiles: [URL] = []
 
-    let enumerator = self.enumerator(atPath: baseURL.path)
-    while let path = enumerator?.nextObject() as? String {
-      let fileURL = URL(string: path, relativeTo: baseURL)
-      if let fileURL = fileURL, fileURL.pathExtension == "swift" {
```
bachand: What about this didn't work? In any case, the new approach seems a bit cleaner.

fdiaz: Apparently it's a known bug. I discovered this by using this script and then confirmed with the unit test.

PR #7, opened 2020-03-25
Reorganize project and co-locate test targets with implementation

fdiaz:
This reorganizes the project so we have three main folders:

- SwiftInspector
- SwiftInspectorCommands
- SwiftInspectorCore

This also co-locates the Test targets with the target is being tested instead of on a separate folder. This makes it easy to discover these tests and update them.

This helps us since code coverage is not picked up for the main executable, so by moving the commands to a library we can have better code coverage information.

image


Tests/LinuxMain.swift
```diff
@@ -1,7 +0,0 @@
-import XCTest
-
-import SwiftInspectorTests
-
-var tests = [XCTestCaseEntry]()
-tests += SwiftInspectorTests.allTests()
-XCTMain(tests)
```
fdiaz: We're not using this file, and if we needed to (in order to run this project in linux) we should generate this in CI instead of manually maintaining this.

bachand:
I think it could still be useful to have a test for the executable that actually runs the built executable.

When you generate a SPM CLI project by default, they give you that test for free.

Even if it's not in code coverage I think it's a nice sanity check that everything integrated together well.

Tests/LinuxMain.swift
bachand: Why do you think we should generate it in CI?

fdiaz:
@bachand all the `SwiftInspectorCommand` test indeed run the executable 😁

bachand:
@fdiaz how does the test for the `SwiftInspectorCommands` target run the executable, since that target doesn't actually build the full binary? Maybe they work because you build everything (which includes the full binary) and _then_ run the tests for each target?

I'm having trouble understanding how this would work if you cleaned and then just ran the tests for `SwiftInspectorCommands`. It feels like that would fail?

fdiaz:
> @fdiaz how does the test for the SwiftInspectorCommands target run the executable, since that target doesn't actually build the full binary? Maybe they work because you build everything (which includes the full binary) and then run the tests for each target?

@bachand we don't test individual targets, they're all bundled together in the executable scheme and in the package scheme which is what we use. You're correct that if we didn't build the executable target before testing  this would fail, but there's no reason why we wouldn't do that so I'm not worried about it honestly.

Tests/LinuxMain.swift
bachand: @fdiaz ^

bachand:
OK I’m glad I’m not crazy in thinking that there is a _potential_ issue here, and I’m fine if we don’t think it’s important to address since it seems to have no material implications at this time 👍

PR #10, opened 2020-03-26
Add typealias command

fdiaz:
**This PR is easier to review commit-by-commit**

This adds a new `typealias` command that:
- Finds all the typealias declarations in a file or folder and outputs all the names
- Has an optional `--name`  parameter to filter for a single typealias and also outputs information related to the identifiers of the typealias

### DIFF
--- a/Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
@@ -1,0 +1,85 @@
+// Created by Francisco Diaz on 3/25/20.
+//
+// Copyright (c) 2020 Francisco Diaz
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import Nimble
+import Quick
+@testable import SwiftInspectorCore
+
+final class TypealiasCommandSpec: QuickSpec {
+  override func spec() {
+    var fileURL: URL!
+    var path: String!
+
+    beforeEach {
+      fileURL = try? Temporary.makeFile(content: "final class Some: SomeType { }")
+      path = fileURL?.path ?? ""
+    }
+
+    afterEach {
+      try? Temporary.removeItem(at: fileURL)
+    }
+
+    describe("run") {
+      context("with no arguments") {
+        it("fails") {
+          let result = try? TestTask.run(withArguments: ["typealias"])
+          expect(result?.didFail) == true
+        }
+      }
+
+      context("with no --name argument") {
+        it("fails") {
+          let result = try? TestTask.run(withArguments: ["typealias", "--path", path])
+          expect(result?.didFail) == true
+        }
+      }
+
+      context("with an empty --name argument") {
+        it("fails") {
+          let result = try? TestTask.run(withArguments: ["typealias", "--name", "", "--path", path])
+          expect(result?.didFail) == true
+        }
+      }
+
+      context("with all arguments") {
+        context("when path doesn't exist") {
+          it("fails") {
+            let result = try? TestTask.run(withArguments: ["typealias", "--name", "SomeAlias", "--path", "/abc"])
+            expect(result?.didSucceed) == false
+          }
+        }
+
+        context("when path exists") {
+          it("succeeds") {
+            let result = try? TestTask.run(withArguments: ["typealias", "--name", "SomeAlias", "--path", path])
+            expect(result?.didSucceed) == true
+          }
+        }
+
+      }
+
+    }
+  }
+}
--- a/Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
@@ -33,7 +33,7 @@ final class TypealiasCommandSpec: QuickSpec {
     var path: String!
 
     beforeEach {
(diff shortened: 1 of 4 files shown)


Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    afterEach {
+      try? Temporary.removeItem(at: fileURL)
+    }
+
+    describe("run") {
+      context("with no arguments") {
+        it("fails") {
+          let result = try? TestTask.run(withArguments: ["typealias"])
```
bachand: I'm a bit nervous about how this actually works (see https://github.com/fdiaz/SwiftInspector/pull/7#issuecomment-604760576). It seems like it would fail under these circumstances, since the binary wouldn't exist.

1. Clean
2. Only build `SwiftInspectorCommands`
3. Run tests for `SwiftInspectorCommands`

But we can dig into this separately.

Sources/SwiftInspectorCore/Tests/TypealiasAnalyzerSpec.swift
```diff
@@ -0,0 +1,109 @@
...
+          expect(result?.first?.identifiers) == ["TypeA", "TypeB"]
+        }
+      }
+
+      context("with multiple typealias statements") {
+        beforeEach {
+          fileURL = try! Temporary.makeFile(
+            content: """
+                     final class Some {
+                       typealias SomeTypealias = SomeType & SomeOtherType
```
bachand: Can we add a test for when it's across multiple lines?

```swift
typealias SomeTypealias = SomeType &
  SomeOtherType
```

Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
```diff
@@ -50,24 +50,44 @@ final class TypealiasCommandSpec: QuickSpec {
...
+        }
+
+        it("does not output the typelias identifiers") {
+          expect(result?.outputMessage).toNot(contain("SomeType"))
         }
       }
 
       context("with an empty --name argument") {
-        it("fails") {
+        it("succeeds") {
```
bachand: Do we really want to support this? Passing an empty name feels like misuse? But I guess maybe the argument for doing this would be that it's empty if you don't specify it?

Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
```diff
@@ -50,24 +50,44 @@ final class TypealiasCommandSpec: QuickSpec {
...
+      }
+
+      context("with an empty --path argument") {
+        it("fails") {
+          let result = try? TestTask.run(withArguments: ["typealias", "--path", ""])
           expect(result?.didFail) == true
         }
       }
 
       context("with all arguments") {
```
bachand: I think it would be nice to be more specific to future proof this. If you add a new argument you may forget to update this. Just a thought.

Sources/SwiftInspectorCommands/TypealiasCommand.swift
```diff
@@ -24,33 +24,71 @@
...
+  private func outputString(from statement: TypealiasStatement) -> String {
+    guard !name.isEmpty else {
+      return statement.name
+    }
+
+    return "\(statement.name) \(statement.identifiers.joined(separator: " "))"
+  }
+}
+
+private struct Help {
```
bachand: I think an `enum` would be a more appropriate type for a namespace like this.

Sources/SwiftInspectorCommands/TypealiasCommand.swift
```diff
@@ -24,33 +24,71 @@
...
     }
   }
+
+  /// Filters the output based on command line inputs
+  private func filterOutput(_ output: [TypealiasStatement]) -> [TypealiasStatement] {
+    guard !name.isEmpty else {
+      return output
+    }
+
+    return output.filter { $0.name == name }
```
bachand: Should we add a test to handle the case where the same typealias occurs twice in the code you're scanning? This could happen if you are scanning multiple modules of if they have different access control.

Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
bachand: Ah I see below.

Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
fdiaz: Answered there 😉

Sources/SwiftInspectorCommands/TypealiasCommand.swift
fdiaz: I've never been fully sold on the "use enums because they can't be initialized argument. For me it. sounds more natural to have a struct with properties than an enum with properties.

Sources/SwiftInspectorCore/Tests/TypealiasAnalyzerSpec.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/10/commits/dc51dcba4124c745d567d7785379f6e5f868cf78

Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/10/commits/dc51dcba4124c745d567d7785379f6e5f868cf78

Sources/SwiftInspectorCommands/TypealiasCommand.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/10/commits/dc51dcba4124c745d567d7785379f6e5f868cf78

Sources/SwiftInspectorCommands/TypealiasCommand.swift
bachand: Makes sense. Since the Airbnb style guide doesn't apply to this repo, I have no issues.

Sources/SwiftInspectorCommands/Tests/TypealiasCommandSpec.swift
```diff
@@ -66,6 +66,20 @@ final class TypealiasCommandSpec: QuickSpec {
...
+          it("only returns one the typealias name once") {
+            fileURL = try? Temporary.makeFile(content: """
+                                                       typealias SomeAlias = SomeType
+                                                       typealias SomeAlias = SomethingElse
+                                                       """
+                                                       )
+            result = try? TestTask.run(withArguments: ["typealias", "--path", fileURL!.path])
+
+            let lines = result?.outputMessage?.split { $0.isNewline }.count
+            expect(lines) == 1
```
bachand: Huh this isn't what I expected. This command can be run on multiple files right? And there is no guarantee that those multiple files are in the same module.

So let's say a file in module A defined `typealias SomeAlias = SomeType` and a file in module B defined `typealias SomeAlias = SomethingElse`. It seems like SwiftInspector _should_ return both.

bachand: Related: let's say you had code like this:

```swift
typealias Foo = Bar

struct MyNamespace {

  typealias Foo = Bar

}
```

I would also expect SwiftInspect to return this twice, or at least print out the information for each of them when I pass `--name Foo`.

How can we handle this valid syntax better?

fdiaz: Oh! You're right. So for this case, it shouldn't since this will only output the typealias name, so IMHO we shouldn't repeat the same output and we should only pass it back once (as we do with the imports statements)

You're right that when passing --name we should return both. I'll add that!

fdiaz: I'd say if they're the same it'd be expected to pass the information back once, if we have something different with the same name though and a `--name` parameter I agree that we should pass it back twice.

PR #11, opened 2020-03-27
Add tests for multiple typealias with the same name

fdiaz:
Add tests to cover for the cases @bachand pointed out here https://github.com/fdiaz/SwiftInspector/pull/10#discussion_r399408837

### DIFF
--- a/Sources/SwiftInspectorCore/Tests/TypealiasAnalyzerSpec.swift
+++ b/Sources/SwiftInspectorCore/Tests/TypealiasAnalyzerSpec.swift
@@ -105,6 +105,36 @@ final class TypealiasAnalyzerSpec: QuickSpec {
         }
       }
 
+      context("with multiple typealias with the same name") {
+        beforeEach {
+          fileURL = try! Temporary.makeFile(
+            content: """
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorCore/Tests/TypealiasAnalyzerSpec.swift
```diff
@@ -105,6 +105,36 @@ final class TypealiasAnalyzerSpec: QuickSpec {
...
+        }
+
+        it("returns all the typealias identifiers") {
+          let result = try? sut.analyze(fileURL: fileURL)
+          expect(result?.count) == 2
+          expect(result?.first?.identifiers) == ["Bar"]
+          expect(result?.last?.identifiers) == ["Bar"]
+        }
+      }
+
```
fdiaz: My philosophy on these type of cases is that the Analyzer should return everything it "sees" and then the Command should be in charge of filtering the output to be consumed by the user. That's why in here we return twice the same information, but in the Command we actually filter to display it to the user once since duplicating this information doesn't provide value.

PR #13, opened 2020-03-29
Add type location command

bachand:
This PR is best reviewed in files changed view.

## The change

This command allows you to find the lines of a file that contain a type declaration.

```
OVERVIEW: Finds the line numbers on which a type is declared

USAGE: SwiftInspector type-location --name <name> --path <path>

OPTIONS:
  --name <name>           The name of the type to find the location of 
        This may be a enum, class, struct, or protocol.
  --path <path>           The absolute path to the file to inspect 
```

This command can be useful if you are trying to move code. It allows you to find a type definition within a file and extract it easily.

The command returns the start and end index of the type. So if your type starts on line 2 (0 indexed) and ends on line 4, you will get back.

```
$ swiftinspector --name Foo --path /MyFile.swift
2 4
```

If multiple types with this name are found, you will get back one line for each type.

```
$ swiftinspector --name Foo --path /MyFileWithMultipleOfTheSameType.swift
2 4
10 13
```

## Testing

I wrote lots of tests. Also, if you run this command on the code I wrote, it works. Remember that it returns 0 indexed line numbers!

(I checked that all of these were correct when I ran them. If the code changes as this PR evolves they may not match up to this PR)

```shell
$ SwiftInspector type-location --name TypeLocationCommand --path /Users/me/repos/SwiftInspector/Sources/SwiftInspectorCommands/TypeLocationCommand.swift
28 70
```

```shell
$ SwiftInspector type-location --name TypeLocationAnalyzer --path /Users/me/repos/SwiftInspector/Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift 
27 53
```

```shell
$ SwiftInspector type-location --name TypeLocationSyntaxVisitor --path /Users/me/repos/SwiftInspector/Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift 
57 130
```

```shell
$ SwiftInspector type-location --name LocatedType --path /Users/me/repos/SwiftInspector/Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift 
135 142
```

## Future work

It would be cool to have a way to find extension of a type too.

Could also be useful to have a way to include documentation comments above a type.

### DIFF
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -83,18 +83,18 @@ final class TypeLocationCommandSpec: QuickSpec {
               expect(result?.didFail) == true
             }
           }
-        }
 
-        context("when path is a directory") {
-          beforeEach { pathURL = try? Temporary.makeFolder() }
+          context("path is a directory") {
+            beforeEach { pathURL = try? Temporary.makeFolder() }
 
-          it("fails") {
-            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
-            expect(result?.didFail) == true
+            it("fails") {
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
+              expect(result?.didFail) == true
+            }
           }
         }
 
-        context("when path is a file") {
+        context("all arguments are valid") {
           beforeEach { pathURL = try? Temporary.makeFile(content: "struct Foo { }") }
 
           it("succeeds") {
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -111,7 +111,7 @@ final class TypeLocationCommandSpec: QuickSpec {
               pathURL = try? Temporary.makeFile(content: contents)
 
               let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
-              expect(result?.outputMessage) == "2 2"
+              expect(result?.outputMessage).to(contain("2 2"))
             }
           }
 
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -26,6 +26,7 @@ import Nimble
 import Quick
 import Foundation
 
+@testable import SwiftInspectorCommands
 @testable import SwiftInspectorCore
 
 final class TypeLocationCommandSpec: QuickSpec {
@@ -33,81 +34,95 @@ final class TypeLocationCommandSpec: QuickSpec {
   override func spec() {
     var pathURL: URL!
 
-    describe("run") {
+    describe("TypeLocationCommand") {
 
-      context("with no arguments") {
-        it("fails") {
-          let result = try? TestTypeLocationTask.run(path: nil, name: nil)
-          expect(result?.didFail) == true
-        }
-      }
+      describe("run") {
 
-      context("path is valid") {
-        beforeEach { pathURL = try? Temporary.makeFile(content: "struct Foo { }") }
-        afterEach { try? Temporary.removeItem(at: pathURL) }
-
-        context("with no --name argument") {
+        context("with no arguments") {
           it("fails") {
-            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: nil)
+            let result = try? TestTypeLocationTask.run(path: nil, name: nil)
             expect(result?.didFail) == true
           }
         }
 
-        context("with an empty --name argument") {
+        context("path is valid") {
+          beforeEach { pathURL = try? Temporary.makeFile(content: "struct Foo { }") }
+          afterEach { try? Temporary.removeItem(at: pathURL) }
+
+          context("with no --name argument") {
+            it("fails") {
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: nil)
+              expect(result?.didFail) == true
+            }
+          }
+
+          context("with an empty --name argument") {
+            it("fails") {
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "")
+              expect(result?.didFail) == true
+            }
+          }
+        }
+
+        context("name is valid") {
+          context("with no --path argument") {
+            it("fails") {
+              let result = try? TestTypeLocationTask.run(path: nil, name: "Foo")
+              expect(result?.didFail) == true
+            }
+          }
+
+          context("with an empty --path argument") {
+            it("fails") {
+              let result = try? TestTypeLocationTask.run(path: "", name: "Foo")
+              expect(result?.didFail) == true
+            }
+          }
+        }
+
+        context("when path is a directory") {
+          beforeEach { pathURL = try? Temporary.makeFolder() }
+          afterEach { try? Temporary.removeItem(at: pathURL) }
+
           it("fails") {
-            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "")
+            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
             expect(result?.didFail) == true
           }
         }
-      }
 
-      context("name is valid") {
-        context("with no --path argument") {
-          it("fails") {
-            let result = try? TestTypeLocationTask.run(path: nil, name: "Foo")
-            expect(result?.didFail) == true
+        context("when path is a file") {
+          beforeEach { pathURL = try? Temporary.makeFile(content: "struct Foo { }") }
+          afterEach { try? Temporary.removeItem(at: pathURL) }
+
+          it("succeeds") {
+            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
+            expect(result?.didSucceed) == true
+          }
+
+          it("outputs the correct line numbers") {
+            let contents =
+            """
+            import Foundation
+
+            struct Foo { }
+            """
+
+            pathURL = try? Temporary.makeFile(content: contents)
+
+            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
+            expect(result?.outputMessage) == "2 2"
           }
         }
-
-        context("with an empty --path argument") {
-          it("fails") {
-            let result = try? TestTypeLocationTask.run(path: "", name: "Foo")
-            expect(result?.didFail) == true
-          }
-        }
       }
+    }
 
-      context("when path is a directory") {
-        beforeEach { pathURL = try? Temporary.makeFolder() }
-        afterEach { try? Temporary.removeItem(at: pathURL) }
+    describe("TypeLocation") {
 
-        it("fails") {
-          let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
-          expect(result?.didFail) == true
-        }
-      }
+      describe("outputString") {
 
-      context("when path is a file") {
-        beforeEach { pathURL = try? Temporary.makeFile(content: "struct Foo { }") }
-        afterEach { try? Temporary.removeItem(at: pathURL) }
-
-        it("succeeds") {
-          let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
-          expect(result?.didSucceed) == true
-        }
-
-        it("outputs the correct line numbers") {
-          let contents =
-          """
-          import Foundation
-
-          struct Foo { }
-          """
-
-          pathURL = try? Temporary.makeFile(content: contents)
-
-          let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
-          expect(result?.outputMessage) == "2 2"
+        it("shows index of start and end line") {
+          let typeLocation = TypeLocation(indexOfStartingLine: 10, indexOfEndingLine: 12)
+          expect(typeLocation.outputString()) == "10 12"
         }
       }
     }
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -130,6 +130,29 @@ final class TypeLocationCommandSpec: QuickSpec {
               expect(result?.outputMessage).to(equal("\n"))
             }
           }
+
+          context("multiple types found") {
+            it("outputs line numbers for each type") {
+              let contents =
+              """
+              import Foundation
+
+              struct Foo { }
+
+              public final class Bar {
+                enum Foo { }
+              }
+              """
+
+              pathURL = try? Temporary.makeFile(content: contents)
+
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
+              let lines = result?.outputMessage?.split { $0.isNewline }
+              expect(lines?.count) == 2
+              expect(lines?.first).to(equal("2 2"))
+              expect(lines?.last).to(equal("5 5"))
+            }
+          }
         }
       }
     }
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -158,18 +158,6 @@ final class TypeLocationCommandSpec: QuickSpec {
           }
         }
       }
-
-      describe("outputString") {
-
-        it("shows index of start and end line") {
-          let sut = TypeLocationCommand()
-          let locatedType = LocatedType(
-            name: "MyType",
-            indexOfStartingLine: 10,
-            indexOfEndingLine: 12)
-          expect(sut.outputString(from: locatedType)) == "10 12"
-        }
-      }
     }
   }
 }
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -99,18 +99,36 @@ final class TypeLocationCommandSpec: QuickSpec {
             expect(result?.didSucceed) == true
           }
 
-          it("outputs the correct line numbers") {
-            let contents =
-            """
-            import Foundation
+          context("type is found") {
+            it("outputs the correct line numbers") {
+              let contents =
+              """
+              import Foundation
 
-            struct Foo { }
-            """
+              struct Foo { }
+              """
 
-            pathURL = try? Temporary.makeFile(content: contents)
+              pathURL = try? Temporary.makeFile(content: contents)
 
-            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
-            expect(result?.outputMessage) == "2 2"
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
+              expect(result?.outputMessage) == "2 2"
+            }
+          }
+
+          context("type is not found") {
+            it("outputs nothing") {
+              let contents =
+              """
+              import Foundation
+
+              struct Foo { }
+              """
+
+              pathURL = try? Temporary.makeFile(content: contents)
+
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Bar")
+              expect(result?.outputMessage).to(equal("\n"))
+            }
           }
         }
       }
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -158,18 +158,16 @@ final class TypeLocationCommandSpec: QuickSpec {
           }
         }
       }
-    }
-
-    describe("LocatedType") {
 
       describe("outputString") {
 
         it("shows index of start and end line") {
+          let sut = TypeLocationCommand()
           let locatedType = LocatedType(
             name: "MyType",
             indexOfStartingLine: 10,
             indexOfEndingLine: 12)
-          expect(locatedType.outputString()) == "10 12"
+          expect(sut.outputString(from: locatedType)) == "10 12"
         }
       }
     }
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -130,7 +130,7 @@ final class TypeLocationCommandSpec: QuickSpec {
               pathURL = try? Temporary.makeFile(content: contents)
 
               let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Bar")
-              expect(result?.outputMessage).to(equal("\n"))
+              expect(result?.outputMessage).to(equal(""))
             }
           }
 
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -35,6 +35,12 @@ final class TypeLocationCommandSpec: QuickSpec {
     var pathURL: URL!
 
     describe("TypeLocationCommand") {
+      afterEach {
+        guard let pathURL = pathURL else {
+          return
+        }
+        try? Temporary.removeItem(at: pathURL)
+      }
 
       describe("run") {
 
@@ -47,7 +53,6 @@ final class TypeLocationCommandSpec: QuickSpec {
 
         context("path is valid") {
           beforeEach { pathURL = try? Temporary.makeFile(content: "struct Foo { }") }
-          afterEach { try? Temporary.removeItem(at: pathURL) }
 
           context("with no --name argument") {
             it("fails") {
@@ -82,7 +87,6 @@ final class TypeLocationCommandSpec: QuickSpec {
 
         context("when path is a directory") {
           beforeEach { pathURL = try? Temporary.makeFolder() }
-          afterEach { try? Temporary.removeItem(at: pathURL) }
 
           it("fails") {
             let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
@@ -92,7 +96,6 @@ final class TypeLocationCommandSpec: QuickSpec {
 
         context("when path is a file") {
           beforeEach { pathURL = try? Temporary.makeFile(content: "struct Foo { }") }
-          afterEach { try? Temporary.removeItem(at: pathURL) }
 
           it("succeeds") {
             let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -130,7 +130,7 @@ final class TypeLocationCommandSpec: QuickSpec {
               pathURL = try? Temporary.makeFile(content: contents)
 
               let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Bar")
-              expect(result?.outputMessage).to(equal(""))
+              expect(result?.outputMessage).to(beEmpty())
             }
           }
 
--- a/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
@@ -95,6 +95,20 @@ final class TypeLocationCommandSpec: QuickSpec {
           let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
           expect(result?.didSucceed) == true
         }
+
+        it("outputs the correct line numbers") {
+          let contents =
+          """
+          import Foundation
+
+          struct Foo { }
+          """
+
+          pathURL = try? Temporary.makeFile(content: contents)
+
+          let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
+          expect(result?.outputMessage) == "2 2"
+        }
       }
     }
   }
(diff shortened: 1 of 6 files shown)


Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,186 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Nimble
+import Quick
+import Foundation
+
+@testable import SwiftInspectorCommands
+@testable import SwiftInspectorCore
```
bachand: I think we need to import this to get `Temporary`? Would be cool if we could break that out into another library I think, since I think it would be more nicely separated if the command tests didn't need to import Core.

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,189 @@
...
+
+          context("with an empty --path argument") {
+            it("fails") {
+              let result = try? TestTypeLocationTask.run(path: "", name: "Foo")
+              expect(result?.didFail) == true
+            }
+          }
+        }
+
+        context("when path is a directory") {
```
bachand: I generally think SwiftInspector should be moving this direction and pushing the iteration logic up a level.
I think this keep SwiftInspector focused.

When you have directories you also need to start thinking of how to handle edge cases (e.g. same type, multiple modules). You get a bit of that in a file with nested types, but it's less extreme. I think it's better if those edge cases are pushed to consumers.

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,189 @@
...
+              """
+              import Foundation
+
+              struct Foo { }
+              """
+
+              pathURL = try? Temporary.makeFile(content: contents)
+
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Bar")
+              expect(result?.outputMessage).to(equal("\n"))
```
bachand: I am pretty sure this is what other commands do. Ideally I think we'd print nothing as an empty newline is kind of weird.

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,189 @@
...
+              """
+
+              pathURL = try? Temporary.makeFile(content: contents)
+
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Bar")
+              expect(result?.outputMessage).to(equal(""))
+            }
+          }
+
+          context("multiple types found") {
```
bachand: I knowingly decided not to try to see what happens if a type of the same name is nested within another type of the same name.

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
```diff
@@ -0,0 +1,174 @@
...
+  /// Find the number of newlines within this node.
+  private func countOfNewlines(within node: Syntax) -> UInt {
+    var countOfNewlinesInsideType: UInt = 0
+
+    for (offset, token) in node.tokens.enumerated() {
+      // We've already counted the leading trivia for the first token.
+      if let leadingTrivia = token.leadingTrivia, offset != 0 {
+        countOfNewlinesInsideType += leadingTrivia.countOfNewlines()
+      }
+      if let trailingTrivia = token.trailingTrivia{
```
bachand: I wondered if we needed to not count to trailing trivia for the last token, but it seems to work 🤷‍♂

bachand:
@fdiaz this PR is now ready for review
@joeyg who I think could find this useful for a current project of his

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
```diff
@@ -0,0 +1,174 @@
...
+  private let typeName: String
+}
+
+// TODO: Update to use SyntaxVisitor when this bug is resolved (https://bugs.swift.org/browse/SR-11591)
+private final class TypeLocationSyntaxReader: SyntaxRewriter {
+  init(onNodeVisit: @escaping (_ locatedType: LocatedType) -> Void) {
+    self.onNodeVisit = onNodeVisit
+  }
+
+  override func visitAny(_ node: Syntax) -> Syntax? {
```
joeyg: Is this more efficient than using `  override public func visit(_ node: ClassDeclSyntax) -> DeclSyntax {` and similar methods for the different types?

bachand: @joeyg I started there and then ended up here since it initially made the code a bit easier to follow. Then I made `anyLocatedType`, which I think would allow me to go back to those methods. My decisionmaking wasn't around efficiency.

bachand: From clicking into the `SyntaxRewriter` code it didn't feel like there is a ton of efficiency to gain in that respect. I think where there is more to gain is bailing out of things early or short circuiting the visitations, but I haven't figured out if/how that is possible.

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
fdiaz: There's definitively a tradeoff between making this easier to use and making this easier to maintain. I think most of the use cases for SwiftInspector are actually going through folders, so I'd prefer to allow folders to be traversed for the sake of usability of the tool even if that means maintenance gets more complicated. For this case where we're actually just trying to discover where a specific type is declared I think limiting to the file makes sense though. For something like the imports command that wouldn't make as much sense.

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,189 @@
...
+              """
+              import Foundation
+
+              struct Foo { }
+              """
+
+              pathURL = try? Temporary.makeFile(content: contents)
+
+              let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Bar")
+              expect(result?.outputMessage).to(equal(""))
```
fdiaz: nit: You can also do `to(beEmpty())`

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
```diff
@@ -0,0 +1,80 @@
...
+// SOFTWARE.
+
+import ArgumentParser
+import Foundation
+import SwiftInspectorCore
+
+final class TypeLocationCommand: ParsableCommand {
+  static var configuration = CommandConfiguration(
+    commandName: "type-location",
+    abstract: "Finds the location of a type"
```
fdiaz: nit: "Finds the line numbers where a type is declared"?

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
```diff
@@ -0,0 +1,80 @@
...
+
+extension LocatedType {
+
+  func outputString() -> String {
+    return "\(indexOfStartingLine) \(indexOfEndingLine)"
+  }
+}
+
+// We recognize that a protocol may not strictly be a type, but we're OK with cutting that corner
+// for now...
```
fdiaz: I'm not sure I follow this comment. Can you clarify?

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
```diff
@@ -0,0 +1,80 @@
...
+private let nameArgumentHelp = ArgumentHelp(
```
fdiaz: nit: I don't love top level properties declared like this. Can we namespace this? Let's just follow the Airbnb Style Guide in this repo and use an enum. I can update the usage of a struct in a previous PR. That way, eventually, we can auto-lint this repo by following those rules too.

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
```diff
@@ -0,0 +1,80 @@
...
+      throw InspectorError.emptyArgument(argumentName: "--path")
+    }
+    var isDir: ObjCBool = false
+    guard FileManager.default.fileExists(atPath: path, isDirectory: &isDir) && !isDir.boolValue else {
+      throw InspectorError.invalidArgument(argumentName: "--path", value: path)
+    }
+  }
+}
+
+extension LocatedType {
```
fdiaz: The reason why we're not extending the type that's outputted from the Analyzer in other places is that this won't work as cleanly when we need to have different output types. I actually started with this path and moved away from it when that use case came up for the imports command.

To allow that use case with this pattern, we'd need to pass the output type parameter from the Command to this type. I've been trying to keep a clear separation of types between the Command and the Analyzer types, where the Analyzer types shouldn't know anything "upwards", but it's ok to have the Command know about the Analyzer (downwards) since the Command uses the Analyzer. This would break that contract and would make it somewhat harder to extend when we have more output types.


So the pattern I've been using is to have the command (which already is responsible for input/output) to be the one that manipulates the LocatedType and convert that into output. That way the LocatedType doesn't know anything about commands, but the command keeps being responsible from input/output.

You can read more about [this here](https://github.com/fdiaz/SwiftInspector/blob/master/CONTRIBUTING.md#structure-of-swiftinspector)

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
```diff
@@ -0,0 +1,174 @@
...
+public final class TypeLocationAnalyzer: Analyzer {
+
+  /// - Parameter typeName: The name of the type to locate.
+  /// - Parameter cachedSyntaxTree: The cached syntax tree to return the AST tree from
+  public init(typeName: String, cachedSyntaxTree: CachedSyntaxTree = .init()) {
+    self.typeName = typeName
+    self.cachedSyntaxTree = cachedSyntaxTree
+  }
+
+  /// Analyzes the imports of the Swift file
```
fdiaz: Let's update this comment

Sources/SwiftInspectorCore/Tests/TypeLocationAnalyzerSpec.swift
```diff
@@ -0,0 +1,334 @@
...
+      context("the type is not present") {
+        let content =
+        """
+        import Foundation
+        """
+        fileURL = try? Temporary.makeFile(content: content)
+
+        it("returns nil") {
+          let sut = TypeLocationAnalyzer(typeName: "Foo")
+          let result = try? sut.analyze(fileURL: fileURL).first
```
fdiaz: Should we be checking for an empty array instead here and below where this doesn't find anything?

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
```diff
@@ -0,0 +1,174 @@
...
+// SOFTWARE.
+
+import Foundation
+import SwiftSyntax
+
+// MARK: - TypeLocationAnalyzer
+
+public final class TypeLocationAnalyzer: Analyzer {
+
+  /// - Parameter typeName: The name of the type to locate.
```
fdiaz: nit: 
```suggestion
  /// - Parameter typeName: The name of the type to locate
```

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
fdiaz: Yeah, using SyntaxRewriter here won't give you that, but we can move this to use SyntaxVisitor by using:


```swift
let syntax = try cachedSyntaxTree.syntaxTree(for: fileURL)

var reader = TypeLocationSyntaxReader(shouldVisitIdentifier: shouldVisit) { [unowned self] node in
    ...
}
syntax.walk(&reader)

func shouldVisit(identifier: String) -> Bool {
  ...
}

private final class TypeLocationSyntaxReader: SyntaxVisitor {
  init(
    shouldVisitIdentifier: @escaping (String) -> Bool,
    onNodeVisit: @escaping (SomeNode) -> Void
  )
  {
    self.shouldVisitIdentifier = shouldVisitIdentifier
    self.onNodeVisit = onNodeVisit
  }

  func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
    shouldVisitIdentifier(node.identifier.text) ? .visitChildren : .skipChildren
  }

  func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
    shouldVisitIdentifier(node.identifier.text) ? .visitChildren : .skipChildren
  }

  func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
    shouldVisitIdentifier(node.identifier.text) ? .visitChildren : .skipChildren
  }

  func visit(_ node: SomeNode) -> SyntaxVisitorContinueKind {
    onNodeVisit(node)
    return .visitChildren
  }

  let shouldVisitIdentifier: (String) -> Bool
  let onNodeVisit: (SomeNode) -> Void
}
```

fdiaz: I recently discovered the walk function from @joeyg FWIW 😉

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
```diff
@@ -0,0 +1,174 @@
...
+// MARK: - LocatedType
+
+/// Information about a located type. Indices start with 0.
+public struct LocatedType: Hashable {
+  /// The name of the type.
+  public let name: String
+  /// The first line of the type.
+  public let indexOfStartingLine: UInt
+  /// The last line of the type.
+  public let indexOfEndingLine: UInt
```
fdiaz: Let's use `Int` [here instead](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID320)

> Use UInt only when you specifically need an unsigned integer type with the same size as the platform’s native word size. If this isn’t the case, Int is preferred, even when the values to be stored are known to be nonnegative. A consistent use of Int for integer values aids code interoperability, avoids the need to convert between different number types, and matches integer type inference, as described in Type Safety and Type Inference.

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
bachand: @fdiaz this is aligned with the Airbnb style guide https://github.com/airbnb/swift#private-constants

bachand: And I agree that we should start following the style guide elsewhere 👍

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
bachand: Cool I'll try this. I plan to consider visiting "any" node unless there is feedback that I shouldn't.

fdiaz: @bachand it's probably more performant to not do that and just do processing. in the nodes you care about. That way we need to only move from SwiftSyntax to SwiftInspector for the nodes we care about instead of checking twice for each node type (once in SwiftInspector and once in SwiftSyntax)

bachand: @fdiaz I will give it a try and see where I get

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
bachand: @fdiaz I see how it's useful for some commands. When used at scale, I can also see performance wins. Staying in Swift is more performance than calling into Swift many times from a higher level language like Ruby or Bash. But at the same time, you could solve that problem by passing a file list....

As this tool becomes more mature, I think we'll want to standardize more. But if you're OK with it (which it sounds like you are by this comment) I guess we can have commands handle this differently for now.

Maybe this is one of those things where the answer will become more obvious over time.

bachand: Another thing to consider is _how_ we intend for this to be used. At work, we usually invoke this from a higher level language and no human is running it on the CLI. If you are actually running this yourself on the CLI I think it becomes harder to reason about directories. Not sure exactly what to do with that thought, but something to chew on.

fdiaz: > If you are actually running this yourself on the CLI I think it becomes harder to reason about directories

I feel the opposite. If we don't provide a way to easily run in directories, this tool becomes a lot less useful since you have to manually traverse the directories and run this file on each Swift file. That's not as useful as inspecting the usage of code in your codebase.

bachand: Ah let me clarify. I think we're actually mostly in agreement.

What I meant is that dealing with the _output_ can be hard when you're using this tool manually on the CLI. The reason is that the file paths are long and then the output is hard to visually read. Also, as we discussed, the potential for spaces in the file path makes it hard to use tools like `cut`. This last issue could be mitigated by putting the file path last, or in other ways.

I thought about this some more today. The more I think about it I do see value in supporting traversing folders for as many commands as possible _eventually_. I think the thing we'd need to figure out is how best to display the paths.

I kind of like how `grep` works in this regard. By default `grep` accepts a path.

```shell
grep "myword" ./Foo/Swift
```

But you can also pass `-R` to traverse directories.

```shell
grep "myword" -R .
```

Maybe the file names aren't as big of an issue as I thought. I guess with `grep` I usually pipe to a file and then process the output from there 🤔  It seems reasonable that I would do the same for SwiftInspector output. We could solve the spaces in paths issue in any number of ways, including by adding a way to specify an output separator, which you could set to a tab instead of a space.

Anyway, I think that there is an opportunity to _standardize_ on an approach for this across commands _eventually_.

Honestly though I think a more pressing issue is getting these commands to work with relative paths 😂

bachand: @fdiaz I think we landed in a good place. I'm resolving this conversation. If you think we still have more to discuss, please unresolve or if that's not possible (I forget if it is) add a new comment 👍

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
bachand: https://github.com/fdiaz/SwiftInspector/pull/13/commits/18d41a578373eed011f11f50ff14fa178c8069f6

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
bachand: I updated to something similar https://github.com/fdiaz/SwiftInspector/pull/13/commits/dfb8f58469e6a0ea03e16cfbd813961e9a991d3d

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
bachand: Consider this output. Swift doesn't seem to call a protocol a type.

Screen Shot 2020-04-04 at 2 18 47 PM

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
bachand: I actually do think this is a good rule. I remember working on it years ago. The thinking is that a namespace isn't that valuable when something is private, since it can't collide with other declarations in other file. So if possible we defer to writing less code.

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
bachand: I updated another one while I was at it. https://github.com/fdiaz/SwiftInspector/pull/13/commits/19a9bf772c3ce36580384cddf27224e2edc94309

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
bachand: https://github.com/fdiaz/SwiftInspector/pull/13/commits/3453388c40c16240161f2360a347f9a3f7ac4a82, https://github.com/fdiaz/SwiftInspector/pull/13/commits/0878d64e87c2eb4508a32f601a736ff334548e65

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
fdiaz: @bachand are you referring to `ArgumentHelp`? `ArgumentHelp` it's a struct, not a type.

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
bachand: https://github.com/fdiaz/SwiftInspector/pull/13/commits/4b2f3e3b36d198cf222336fc663836320f289cd6

Sources/SwiftInspectorCore/Tests/TypeLocationAnalyzerSpec.swift
bachand: I originally returned a `LocatedType?` and only later made it an array. I didn't clean up as well as I should have https://github.com/fdiaz/SwiftInspector/pull/13/commits/4c1a086b040821e5bee22a6b8313786207fdf7d6

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
bachand: Ah sorry let me be more clear.

The name of this command is "type location". So it finds the location of _types_.

I am not sure that Swift considers protocols to be "types". In Swift, a structure or class or enum is clearly a type. But I'm not sure the language officially thinks that a protocol is a type.

What I'm trying to communicate with the code output is that invoking `.self` on a protocol functions differently than invoking `.self` on a class, enum, or structure. This output may be more clear.

Screen Shot 2020-04-04 at 3 52 07 PM

So anyway the reason that I added this comment was to explain that even though the command is called "type location" it is able to find protocols too, even if Swift doesn't consider a protocol to actually be a type.

Does that make sense?

Happy to remove or make more clear, if you have suggestions.

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -157,25 +157,6 @@ final class InitializerAnalyzerSpec: QuickSpec {
             ]
           }
         }
-
-        context("with a closure") {
```
bachand, line 161: @fdiaz adding this test seemed like a mistake since it is testing a tuple, just like the one above. Removing in an effort to be helpful but lmk if I missed something.

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
bachand: I moved to `SyntaxVisitor`. I didn't see an easy way to skip children since I need to count the tokens within nodes to determine the current line number.

Sources/SwiftInspectorCore/TypeLocationAnalyzer.swift
bachand: https://github.com/fdiaz/SwiftInspector/commit/f4e34806407bb4959c3cf5489518240c09f90a28

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,187 @@
...
+              expect(lines?.last).to(equal("5 5"))
+            }
+          }
+        }
+      }
+
+      describe("outputString") {
+
+        it("shows index of start and end line") {
+          let sut = TypeLocationCommand()
```
bachand, line 165: I wasn't sure I could instantiate the command due to the property wrappers, but I guess I can!

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
bachand: (I made this comment on the PR too 👍 )

Sources/SwiftInspectorCommands/TypeLocationCommand.swift
fdiaz: @bachand Oh gotcha! I think you're conflating Types with Meta Types. The things you're referring to are meta types, [see this](https://docs.swift.org/swift-book/ReferenceManual/Types.html#ID455).

`Foo` in your example is the type of your protocol, while `Foo.Protocol` is the meta type.

bachand: OK I removed this comment https://github.com/fdiaz/SwiftInspector/pull/13/commits/be073028222b98aba43c0b89d4c8aa243e08f655

bachand:
@fdiaz this PR is ready for re-review. Definitely look at the files changed view 👍

Sources/SwiftInspectorCore/Tests/TypeLocationAnalyzerSpec.swift
```diff
@@ -0,0 +1,405 @@
...
+          let sut = TypeLocationAnalyzer(typeName: "Foo")
+          let result = try? sut.analyze(fileURL: fileURL)
+          let typeLocation = result?.first
+          expect(typeLocation?.indexOfStartingLine) == 2
+          expect(typeLocation?.indexOfEndingLine) == 4
+        }
+      }
+    }
+  }
+}
```
fdiaz, line 405: These were super easy to read! Thank you!

Sources/SwiftInspectorCore/TypealiasAnalyzer.swift
```diff
@@ -32,7 +32,7 @@ public final class TypealiasAnalyzer: Analyzer {
     self.cachedSyntaxTree = cachedSyntaxTree
   }
 
-  /// Analyzes the imports of the Swift file
+  /// Analyzes the matching initializers of the Swift file
```
fdiaz, line 35: Thank you!

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,187 @@
...
+        context("when path is a directory") {
+          beforeEach { pathURL = try? Temporary.makeFolder() }
+
+          it("fails") {
+            let result = try? TestTypeLocationTask.run(path: pathURL.path, name: "Foo")
+            expect(result?.didFail) == true
+          }
+        }
+
+        context("when path is a file") {
```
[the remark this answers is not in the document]
bachand: I was trying to create some separation between the tests that check arguments (which are the ones that you linked to) and the tests that actually test functionality.

I find it a bit easier to read tests when groups of tests are testing the same type of thing.

I wasn't thinking ahead to expanding this to be able to handle directories. I wanted to create a group of tests that are really testing the "meat" of the problem.

Another name for this group of tests could be "when all arguments are valid" but this felt a bit more descriptive 🤷‍♂

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -0,0 +1,187 @@
...
+              let lines = result?.outputMessage?.split { $0.isNewline }
+              expect(lines?.count) == 2
+              expect(lines?.first).to(equal("2 2"))
+              expect(lines?.last).to(equal("5 5"))
+            }
+          }
+        }
+      }
+
+      describe("outputString") {
```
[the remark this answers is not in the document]
bachand: I don't feel strongly. I removed https://github.com/fdiaz/SwiftInspector/pull/13/commits/51041bc608658908c65bdab35dbe1210f0333d66

@fdiaz can you help me understand how it ties us to ArgumentParser though? I guess because I'm instantiating the command directly?

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
```diff
@@ -83,18 +83,18 @@ final class TypeLocationCommandSpec: QuickSpec {
               expect(result?.didFail) == true
             }
           }
-        }
 
-        context("when path is a directory") {
-          beforeEach { pathURL = try? Temporary.makeFolder() }
+          context("path is a directory") {
```
bachand, line 87: Moved under the "name is valid section"

Sources/SwiftInspectorCommands/Tests/TypeLocationCommandSpec.swift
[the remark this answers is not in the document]
bachand: I adjusted https://github.com/fdiaz/SwiftInspector/pull/13/commits/18c29cb8598259b073fe369fe7b0393429baf7d3

bachand:
@fdiaz I don't think this PR is building in CI. I don't see any CI checks running for past commits. Any idea why that would be the case?

bachand:
I enable Github Actions on my fork. Then I followed [this suggestion](https://github.community/t5/GitHub-Actions/Forked-repo-doesn-t-trigger-action/m-p/32575/highlight/true#M1189) and deleted the `ci.yml` file and recreated it.

bachand:
Taking those steps did allow me to get a [build](https://github.com/bachand/SwiftInspector/runs/562602476?check_suite_focus=true) running in my fork.

However, it is not registering with the status checks on this PR.

I'm also confused by the code coverage report. I found a link to it in the build. [This page](https://codecov.io/gh/bachand/SwiftInspector/tree/0b3713584f7a63843a753644a12033677d2c146f/Sources/SwiftInspectorCore) for example seems off since it seems like we must have higher code coverage for the `TypeliasAnalzyer`.

Screen Shot 2020-04-05 at 1 34 26 PM

fdiaz:
@bachand I see CI running fine https://github.com/fdiaz/SwiftInspector/pull/13/checks?check_run_id=562636659

Screen Shot 2020-04-05 at 1 59 08 PM

bachand:
OK wow. Creating another PR with a branch (of the same name) _in this repo_ got a build going that made this PR's CI checks pass.

Code coverage looks good locally.

Screen Shot 2020-04-05 at 1 57 39 PM
Screen Shot 2020-04-05 at 1 57 45 PM

I made an issue https://github.com/fdiaz/SwiftInspector/issues/28

bachand:
(I removed the commits where I was deleting and re-adding the `ci.yml` file since we fixed the issue of pull requests not building in CI https://github.com/fdiaz/SwiftInspector/pull/30)

PR #15, opened 2020-03-30
Add initializer command

fdiaz:
Adds a command to find the initializer arguments given a type name.

```
OVERVIEW: Finds information about the initializers of the specified type

USAGE: SwiftInspector initializer --path <path> --name <name> [--enable-type-only] [--disable-type-only]

OPTIONS:
  --path <path>           The absolute path to the file to inspect 
  --name <name>           The name of the type to look information of the
                          initializer 
  -t, --enable-type-only/--disable-type-only
                          The granularity of the output (default: true)
        Outputs a list of the type names by default. If disabled it outputs the
        name of the parameter and the name of the type (e.g. 'foo,Int
        bar,String')

```

### DIFF
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -35,10 +35,10 @@ final class InitializerCommand: ParsableCommand {
   @Option(help: "The absolute path to the file to inspect")
   var path: String
 
-  @Option(help: "The name of the type to look information of the initializer")
+  @Option(help: "The name of the type whose initializer information we'll be looking for")
   var name: String
 
-  @Flag(name: .shortAndLong, default: true, inversion: .prefixedEnableDisable, help: Help.typeOnly)
+  @Flag(name: .shortAndLong, default: true, inversion: .prefixedEnableDisable, help: typeOnlyHelp)
   var typeOnly: Bool
 
   /// Runs the command
@@ -48,7 +48,7 @@ final class InitializerCommand: ParsableCommand {
     let fileURL = URL(fileURLWithPath: path)
 
     let initializerStatements = try analyzer.analyze(fileURL: fileURL)
-    let outputArray = initializerStatements.map { outputString(from: $0)}
+    let outputArray = initializerStatements.map { outputString(from: $0) }
 
     let output = outputArray.joined(separator: "\n")
     print(output)
@@ -63,7 +63,7 @@ final class InitializerCommand: ParsableCommand {
       throw InspectorError.emptyArgument(argumentName: "--path")
     }
     guard FileManager.default.fileExists(atPath: path) else {
-      throw InspectorError.invalidArgument(argumentName: "--path", value: "options.path")
+      throw InspectorError.invalidArgument(argumentName: "--path", value: path)
     }
   }
 
@@ -76,9 +76,7 @@ final class InitializerCommand: ParsableCommand {
   }
 }
 
-private enum Help {
-  static var typeOnly: ArgumentHelp = ArgumentHelp("The granularity of the output",
-                                                   discussion: """
-                                                    Outputs a list of the type names by default. If disabled it outputs the name of the parameter and the name of the type (e.g. 'foo,Int bar,String')
-                                                    """)
-}
+private var typeOnlyHelp = ArgumentHelp("The granularity of the output",
+                                        discussion: """
+                                        Outputs a list of the type names by default. If disabled it outputs the name of the parameter and the name of the type (e.g. 'foo,Int bar,String')
+                                        """)
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -32,14 +32,26 @@ final class InitializerCommand: ParsableCommand {
     abstract: "Finds information about the initializers of the specified type"
   )
 
-  @Option(help: "The absolute path to the file or directory to inspect")
+  @Option(help: "The absolute path to the file to inspect")
   var path: String
 
   @Option(help: "The name of the type to look information of the initializer")
   var name: String
 
+  @Flag(name: .shortAndLong, default: true, inversion: .prefixedEnableDisable, help: Help.typeOnly)
+  var typeOnly: Bool
+
   /// Runs the command
   func run() throws {
+    let cachedSyntaxTree = CachedSyntaxTree()
+    let analyzer = InitializerAnalyzer(name: name, cachedSyntaxTree: cachedSyntaxTree)
+    let fileURL = URL(fileURLWithPath: path)
+
+    let initializerStatements = try analyzer.analyze(fileURL: fileURL)
+    let outputArray = initializerStatements.map { outputString(from: $0)}
+
+    let output = outputArray.joined(separator: "\n")
+    print(output)
   }
 
   /// Validates if the arguments of this command are valid
@@ -54,4 +66,19 @@ final class InitializerCommand: ParsableCommand {
       throw InspectorError.invalidArgument(argumentName: "--path", value: "options.path")
     }
   }
+
+  private func outputString(from statement: InitializerStatement) -> String {
+    if typeOnly {
+      return statement.parameters.map { $0.typeName }.joined(separator: " ")
+    } else {
+      return statement.parameters.map { "\($0.name),\($0.typeName)" }.joined(separator: " ")
+    }
+  }
+}
+
+private enum Help {
+  static var typeOnly: ArgumentHelp = ArgumentHelp("The granularity of the output",
+                                                   discussion: """
+                                                    Outputs a list of the type names by default. If disabled it outputs the name of the parameter and the name of the type (e.g. 'foo,Int bar,String')
+                                                    """)
 }
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -1,0 +1,25 @@
+// Created by Francisco Diaz on 3/27/20.
+//
+// Copyright (c) 2020 Francisco Diaz
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -22,4 +22,36 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+import ArgumentParser
 import Foundation
+import SwiftInspectorCore
+
+final class InitializerCommand: ParsableCommand {
+  static var configuration = CommandConfiguration(
+    commandName: "initializer",
+    abstract: "Finds information about the initializers of the specified type"
+  )
+
+  @Option(help: "The absolute path to the file or directory to inspect")
+  var path: String
+
+  @Option(help: "The name of the type to look information of the initializer")
+  var name: String
+
+  /// Runs the command
+  func run() throws {
+  }
+
+  /// Validates if the arguments of this command are valid
+  func validate() throws {
+    guard !name.isEmpty else {
+      throw InspectorError.emptyArgument(argumentName: "--name")
+    }
+    guard !path.isEmpty else {
+      throw InspectorError.emptyArgument(argumentName: "--path")
+    }
+    guard FileManager.default.fileExists(atPath: path) else {
+      throw InspectorError.invalidArgument(argumentName: "--path", value: "options.path")
+    }
+  }
+}
--- a/Sources/SwiftInspectorCommands/StaticUsageCommand.swift
+++ b/Sources/SwiftInspectorCommands/StaticUsageCommand.swift
@@ -60,7 +60,7 @@ final class StaticUsageCommand: ParsableCommand {
       throw InspectorError.emptyArgument(argumentName: "--path")
     }
     guard FileManager.default.fileExists(atPath: path) else {
-      throw InspectorError.invalidArgument(argumentName: "--path", value: "options.path")
+      throw InspectorError.invalidArgument(argumentName: "--path", value: path)
     }
   }
 
--- a/Sources/SwiftInspectorCommands/Tests/ImportCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/ImportCommandSpec.swift
@@ -28,9 +28,7 @@ import Foundation
 
 @testable import SwiftInspectorCore
 
-final class InspectorCommandSpec: QuickSpec {
-  private var fileURL: URL!
-
+final class ImportCommandSpec: QuickSpec {
   override func spec() {
     describe("run") {
 
--- a/Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
@@ -55,7 +55,11 @@ final class InitializerCommandSpec: QuickSpec {
         var fileURL: URL!
 
         beforeEach {
-          fileURL = try? Temporary.makeFile(content: "")
+          fileURL = try? Temporary.makeFile(content: """
+          final class Some {
+            init(some: String, someInt: Int) {}
+          }
+          """)
         }
 
         afterEach {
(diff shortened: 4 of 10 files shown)


Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+final class InitializerCommand: ParsableCommand {
+  static var configuration = CommandConfiguration(
+    commandName: "initializer",
+    abstract: "Finds information about the initializers of the specified type"
+  )
+
+  @Option(help: "The absolute path to the file to inspect")
+  var path: String
+
+  @Option(help: "The name of the type to look information of the initializer")
```
bachand: Let's reword this sentence. It's slightly off as-is.

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+  private func outputString(from statement: InitializerStatement) -> String {
+    if typeOnly {
+      return statement.parameters.map { $0.typeName }.joined(separator: " ")
+    } else {
+      return statement.parameters.map { "\($0.name),\($0.typeName)" }.joined(separator: " ")
+    }
+  }
+}
+
+private enum Help {
```
bachand: https://github.com/airbnb/swift#private-constants

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+  var typeOnly: Bool
+
+  /// Runs the command
+  func run() throws {
+    let cachedSyntaxTree = CachedSyntaxTree()
+    let analyzer = InitializerAnalyzer(name: name, cachedSyntaxTree: cachedSyntaxTree)
+    let fileURL = URL(fileURLWithPath: path)
+
+    let initializerStatements = try analyzer.analyze(fileURL: fileURL)
+    let outputArray = initializerStatements.map { outputString(from: $0)}
```
bachand:
```suggestion
    let outputArray = initializerStatements.map { outputString(from: $0) }
```

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    let output = outputArray.joined(separator: "\n")
```
bachand: I think it's nice when it prints nothing when it doesn't find anything. Can we do something like this? https://github.com/fdiaz/SwiftInspector/blob/0ac2841c36483d2097e9ed7a15db9a235d9c2bcf/Sources/SwiftInspectorCommands/TypeLocationCommand.swift#L49-L51

bachand: As is I think it prints a newline.

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+
+  /// Validates if the arguments of this command are valid
+  func validate() throws {
+    guard !name.isEmpty else {
+      throw InspectorError.emptyArgument(argumentName: "--name")
+    }
+    guard !path.isEmpty else {
+      throw InspectorError.emptyArgument(argumentName: "--path")
+    }
+    guard FileManager.default.fileExists(atPath: path) else {
```
bachand: Your help documentation says it expects a file. If so, can we verify this is a file? https://github.com/fdiaz/SwiftInspector/blob/0ac2841c36483d2097e9ed7a15db9a235d9c2bcf/Sources/SwiftInspectorCommands/TypeLocationCommand.swift#L62-L65

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+      throw InspectorError.invalidArgument(argumentName: "--path", value: "options.path")
```
bachand: I meant to ask you about this. Isn't this what we want?
```suggestion
      throw InspectorError.invalidArgument(argumentName: "--path", value: path)
```

If not I had trouble following

Sources/SwiftInspectorCommands/Tests/ImportCommandSpec.swift
```diff
@@ -28,9 +28,7 @@ import Foundation
 
 @testable import SwiftInspectorCore
 
-final class InspectorCommandSpec: QuickSpec {
```
bachand: 👍

Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
```diff
@@ -0,0 +1,92 @@
...
+    describe("run") {
+
+      context("with no arguments") {
+        it("fails") {
+          let result = try? TestTask.run(withArguments: ["initializer"])
+          expect(result?.didFail) == true
+        }
+      }
+
+      context("when path is invalid") {
```
bachand: I can start adding "when" and words like that if you think that's best. It does make it more readable.

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -0,0 +1,182 @@
...
+
+  public func analyze(fileURL: URL) throws -> [InitializerStatement] {
+    var statements: [InitializerStatement] = []
+    let syntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
+
+    var reader = InitializerSyntaxReader(shouldVisitIdentifier: shouldVisit) { [unowned self] node in
+      let statement = self.initializerStatement(from: node)
+      statements.append(statement)
+    }
+    syntax.walk(&reader)
```
bachand: OK I need to adopt this

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -0,0 +1,182 @@
...
+
+  private func initializerStatement(from node: InitializerDeclSyntax) -> InitializerStatement {
+    let parameters = findParameters(from: node)
+    let modifiers = findModifiers(from: node)
+    return InitializerStatement(typeName: name, parameters: parameters, modifiers: modifiers)
+  }
+
+  private func findParameters(from node: InitializerDeclSyntax) -> [InitializerStatement.Parameter] {
+    let functionList = node.children
+      .compactMap { $0 as? ParameterClauseSyntax }
```
bachand: Would it make sense to assert there is only one, to sanity check?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -0,0 +1,182 @@
...
+      .first?.children
+      .compactMap { $0 as? FunctionParameterListSyntax }
+      .first
```
bachand: Same here

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -0,0 +1,182 @@
...
+      let parameter = findParameter(from: parameterNode)
+      parameters.append(parameter)
+    }
+
+    return parameters
+  }
+
+  private func findParameter(from node: FunctionParameterSyntax) -> InitializerStatement.Parameter {
+    let name = node.firstName?.text ?? ""
+    let typeName = node.children.compactMap { $0 as? SimpleTypeIdentifierSyntax }.first?.name.text ?? ""
```
bachand: What are the other options besides `SimpleTypeIdentifierSyntax`?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -0,0 +1,182 @@
...
+    public let typeName: String
+  }
+
+  public struct Modifier: Equatable, OptionSet  {
+    public let rawValue: Int
+
+    public static let designated = Modifier(rawValue: 1 << 0)
+    public static let convenience = Modifier(rawValue: 1 << 1)
+    public static let override = Modifier(rawValue: 1 << 2)
+    public static let required = Modifier(rawValue: 1 << 3)
```
bachand: Go has a really cool way to do this. Check out `iota`

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -0,0 +1,135 @@
...
+      try? Temporary.removeItem(at: fileURL)
+    }
+
+    describe("analyze(fileURL:)") {
+      context("when there is no initializer") {
+        beforeEach {
+          let content = """
+                        public final class FakeType {}
+                        """
+          fileURL = try? Temporary.makeFile(content: content, name: "ABC")
```
bachand:
```suggestion
          fileURL = try? Temporary.makeFile(content: content)
```
?

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -0,0 +1,135 @@
...
+      }
+
+      context("when there is an initializer") {
+        beforeEach {
+          let content = """
+          public final class FakeType {
+            convenience init(someString: String, someInt: Int) {}
+          }
+          """
+          fileURL = try? Temporary.makeFile(content: content, name: "ABC")
```
bachand:
```suggestion
          fileURL = try? Temporary.makeFile(content: content)
```

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -0,0 +1,135 @@
...
+          public struct AnotherType {
+            init(fooBar: String)
+          }
+
+          public final class FakeType {
+            init(someString: String) {}
+            override init(someString: String, someInt: Int) {}
+          }
+          """
+          fileURL = try? Temporary.makeFile(content: content, name: "ABC")
```
bachand:
```suggestion
          fileURL = try? Temporary.makeFile(content: content)
```

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -0,0 +1,135 @@
...
+
+          expect(result).to(beEmpty())
+        }
+      }
+
+      context("when there is an initializer") {
+        beforeEach {
+          let content = """
+          public final class FakeType {
+            convenience init(someString: String, someInt: Int) {}
```
bachand: Can we add a test for an initializer that spans multiple lines?

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -0,0 +1,135 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Nimble
+import Quick
+import Foundation
+
+@testable import SwiftInspectorCore
+
+final class InitializerAnalyzerSpec: QuickSpec {
```
bachand: What is supposed to happen when the same type exists in multiple places in a file?

```swift
class Foo {
  init(myParam: String) { }
}

struct Bar {
  class Foo {
    init(anotherParam: Int) { }
  }
}
```

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
bachand: Can we also add a test for an initializer that has a default value for one or more parameters?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
fdiaz: I don't think anybody would be hitting that assertion. Remember this code is only used by running the binary or when we're developing so that assertion wouldn't be hit unless we add a test, or we can fail if that happens but that feels to harsh for the end-user.
If we think there's a use case for when we can have multiple `ParameterClauseSyntax` for one initializer I can add. a test but I couldn't come up with a use case.

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
fdiaz: Same as https://github.com/fdiaz/SwiftInspector/pull/15#discussion_r400610689

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
fdiaz: `children` is just an arbitrary list of child `SyntaxChildren`. We don't get a nice list of strongly typed values here so we need to cast for the thing we're looking for. In this case, the "type name" is wrapped in a `SimpleTypeIdentifierSyntax` which is why we're only casting for that when looking for the type.

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
fdiaz: I don't think we need to. This relies on SwiftSyntax doing that for us, so we could add tests like that but IMO we're testing there `SwiftSyntax` ore than we're testing `SwiftInspector` which IMO we shouldn't be testing.

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
fdiaz: We should be returning both initializers. Let me add a test for that.

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: I haven't consider this. Do you  think it's problematic to print a newline? I'd prefer to keep the code simple if we can but if this might become troublesome I think we should remove it.

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: Good call. I'll create a new convenience method to check if there's a swift file at the location 👍

fdiaz: https://github.com/fdiaz/SwiftInspector/issues/17

Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
fdiaz: I don't feel strongly on either direction

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/15/commits/21d3b9d79d11f6aa134bf13a106ab4fc20f77262

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/15/commits/21d3b9d79d11f6aa134bf13a106ab4fc20f77262

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/15/commits/21d3b9d79d11f6aa134bf13a106ab4fc20f77262

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/15/commits/21d3b9d79d11f6aa134bf13a106ab4fc20f77262

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/15/commits/21d3b9d79d11f6aa134bf13a106ab4fc20f77262

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
fdiaz: Right now we'd be only returning the top-level one actually. I'm not sure what's more desirable. Maybe we should create an option to look for the scoped ones? Let's discuss and add this in another PR before we ship a new version of this.

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
bachand: When we use this at work, do we build as Release or Debug?

At home I have an alias shell to run the debug version, so I could hit an assertion locally.

```shell
alias si="$HOME/repos/SwiftInspector/.build/debug/SwiftInspector"
``` 

I don't feel super strongly, though I see little downside in being generous with assertions for things that we truly think can never be hit. Even if you normally are using a Release build where they don't fire, you can switch to a Debug build if you are running into an unexpected issue. Then in that Debug build you have more insight into whether anything odd is happening, which can save debugging time.

And even if they never file, they can be a good way to signal developer intent. Reading the code right now it's not immediately clear if I should expect that there are multiple and we're picking the first, or that there should only ever be one. An assertion would be a way to communicate to a future maintainer that the latter is the case, and it's preferable to a comment IMHO since it does have meaning in the code (under some build configurations).

Anyway the length of this comment doesn't indicate that I feel super strongly. Mostly want to communicate my thinking 👍

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/15/commits/b4d5a0aabebd195ed0ba72f2894a1bd86cdfddeb

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/issues/18

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
bachand: (Though I may not be using this to find initializer parameters at home, so I guess it's questionable if this would be hit 😂)

Sources/SwiftInspectorCommands/InitializerCommand.swift
bachand: I don't think it's terrible but it's not idiomatic. Sibling CLI tools that find stuff (`grep`, `sed`, `find`) are all silent when they don't find anything. I think there is some value in following their lead.

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
bachand: I did think about this when I added my comment. I'm open to that idea. I guess this is where we gain a lot vs. using regexes 💪

Sources/SwiftInspectorCommands/StaticUsageCommand.swift
```diff
@@ -60,7 +60,7 @@ final class StaticUsageCommand: ParsableCommand {
       throw InspectorError.emptyArgument(argumentName: "--path")
     }
     guard FileManager.default.fileExists(atPath: path) else {
-      throw InspectorError.invalidArgument(argumentName: "--path", value: "options.path")
```
bachand: Thanks for fixing across all commamnds!

Issue #17, opened 2020-03-31
Add a convenience method on FileManager to check for a Swift file


fdiaz:
Related to: https://github.com/fdiaz/SwiftInspector/pull/15#discussion_r400585144

PR #19, opened 2020-03-31
Add a helper to check if a URL is a Swift file

fdiaz:
Closes https://github.com/fdiaz/SwiftInspector/issues/17

### DIFF
--- a/Sources/SwiftInspectorCore/FileManager+SwiftFiles.swift
+++ b/Sources/SwiftInspectorCore/FileManager+SwiftFiles.swift
@@ -55,4 +55,12 @@ extension FileManager {
 
     return swiftFiles
   }
+
+  /// Checks whether a URL contains a Swift file
+  ///
+  /// - Parameter baseURL: The URL to look for a Swift file
+  /// - Returns: `true` if the URL contains a Swift file, `false` otherwise
+  public func isSwiftFile(at baseURL: URL) -> Bool {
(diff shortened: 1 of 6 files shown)


Sources/SwiftInspectorCore/FileManager+SwiftFiles.swift
```diff
@@ -55,4 +55,12 @@ extension FileManager {
 
     return swiftFiles
   }
+
+  /// Checks whether a URL contains a Swift file
+  ///
+  /// - Parameter baseURL: The URL to look for a Swift file
+  /// - Returns: `true` if the URL contains a Swift file, `false` otherwise
+  public func isSwiftFile(at baseURL: URL) -> Bool {
+    FileManager.default.fileExists(atPath: baseURL.path) && !baseURL.hasDirectoryPath && baseURL.pathExtension == "swift"
```
bachand: FWIW I suspect that actually checking if it's a directory with the file manager is more robust.
I bet that if you have a directory named `/files/MyDirectory.swift` this would succeed erroneously.
Could be worth adding a test, but depends on how paranoid we want to be.

fdiaz: Good call! it actually works! See https://github.com/fdiaz/SwiftInspector/pull/19/commits/2773c52ee6d508bf450a07b7ba6e38ccf942a13f

bachand: Wow crazy! I wonder how it knows.

Issue #20, opened 2020-04-01
Improvements to the initializer command


fdiaz:
- [x] Have a way to filter given an argument name e.g. find an initializer that takes an argument with a given name https://github.com/fdiaz/SwiftInspector/pull/22
- [x] Make sure that we support the case where an initializer has a composition of types e.g. `init(someArgument: Some & SomeOther)` https://github.com/fdiaz/SwiftInspector/pull/21

PR #21, opened 2020-04-01
Handle the case where an initializer has a composition of types

fdiaz:
Work on https://github.com/fdiaz/SwiftInspector/issues/20

This fixes an edge case of when an initializer argument contains a composition of elements.

### DIFF
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -71,9 +71,9 @@ final class InitializerCommand: ParsableCommand {
 
   private func outputString(from statement: InitializerStatement) -> String {
     if typeOnly {
-      return statement.parameters.map { $0.typeName }.joined(separator: " ")
+      return statement.parameters.map { $0.typeNames.joined(separator: " ") }.joined(separator: " ")
     } else {
-      return statement.parameters.map { "\($0.name),\($0.typeName)" }.joined(separator: " ")
+      return statement.parameters.map { "\($0.name),\($0.typeNames.joined(separator: ","))" }.joined(separator: " ")
     }
   }
 }
--- a/Sources/SwiftInspectorCore/InitializerAnalyzer.swift
+++ b/Sources/SwiftInspectorCore/InitializerAnalyzer.swift
@@ -82,9 +82,14 @@ public final class InitializerAnalyzer: Analyzer {
 
   private func findParameter(from node: FunctionParameterSyntax) -> InitializerStatement.Parameter {
     let name = node.firstName?.text ?? ""
-    let typeName = node.children.compactMap { $0 as? SimpleTypeIdentifierSyntax }.first?.name.text ?? ""
 
-    return .init(name: name, typeName: typeName)
+    var typeNames: [String] = []
+    var reader = FunctionParameterReader() { identifierNode in
+      typeNames.append(identifierNode.name.text)
+    }
+    node.walk(&reader)
+
+    return .init(name: name, typeNames: typeNames)
   }
 
   private func findModifiers(from node: InitializerDeclSyntax) -> InitializerStatement.Modifier {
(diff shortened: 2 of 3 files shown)


Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -71,9 +71,9 @@ final class InitializerCommand: ParsableCommand {
 
   private func outputString(from statement: InitializerStatement) -> String {
     if typeOnly {
-      return statement.parameters.map { $0.typeName }.joined(separator: " ")
+      return statement.parameters.map { $0.typeNames.joined(separator: " ") }.joined(separator: " ")
```
bachand, line 74: I'm a bit confused by how you parse the output now. It seems like it would be hard to parse `foo,MyProtocol YourProtocol anotherParam:String`.

Should it be `foo,MyProtocol,YourProtocol anotherParam:String` instead?

bachand, line 74: Also while we are thinking about edge cases like this, what happens if the type is a tuple? Or a closure?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -180,3 +203,17 @@ private final class InitializerSyntaxReader: SyntaxVisitor {
   let shouldVisitIdentifier: (String) -> Bool
   let onNodeVisit: (InitializerDeclSyntax) -> Void
 }
+
+private final class FunctionParameterReader: SyntaxVisitor {
```
bachand, line 207: I like the second syntax visitor approach

Sources/SwiftInspectorCommands/InitializerCommand.swift
bachand: In the case where you exclude parameter names, how do you know which types belong to switch initializer parameter? How do I parse `MyProtocol YouProtocol String`?

fdiaz: @bachand this is for the type only case. So this would output:

`MyProtocol YourProtocol String` in your use case.

[Below](https://github.com/fdiaz/SwiftInspector/pull/21/files#diff-f8fee815716808d2efcdd15e9654898aR76) is the update for multiple types when outputting the argument name. It'd look like:

`foo,MyProtocol,YourProtocol anotherParam,String`

fdiaz: Good call! Let me add more tests and try those use cases.

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
fdiaz: Thanks! Yeah I think it makes it a lot cleaner to follow the code.

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: For a tuple we're handling this correctly. For a closure we're just returning the types on the closure itself but it's somewhat weird of a syntax since it'll just list the types.

I created this https://github.com/fdiaz/SwiftInspector/issues/23 to track this since I think this might be a bigger change.

fdiaz: typle commit https://github.com/fdiaz/SwiftInspector/pull/22/commits/a12e7637a8120e5c9ee2ac0a6761767ae0ebca64

bachand: @fdiaz I'm curious why you chose to output `MyProtocol YourProtocol String` in the type only case vs. `MyProtocol,YourProtocol String`. In any case, I'm fine deferring this until we address https://github.com/fdiaz/SwiftInspector/issues/23

fdiaz: I think its easier to parse when we do that, and if we're saying we only care about the types I don't see a value in adding context for each argument.

What I'm guessing this can be used to is to know what types a certain class takes in case there are types we want to forbid or warn about.

Since we have the `--disable-type-only` flag, if we do care about the parameters and how they're grouped together we can always get that information from there instead.

bachand: > What I'm guessing this can be used to is to know what types a certain class takes in case there are types we want to forbid or warn about.

Makes sense!

PR #22, opened 2020-04-02
Add a way to filter the initializer by argument names

fdiaz:
Fixes https://github.com/fdiaz/SwiftInspector/issues/20

This adds a way to filter the initializer command given a list of argument names

### DIFF
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -41,6 +41,9 @@ final class InitializerCommand: ParsableCommand {
   @Flag(name: .shortAndLong, default: true, inversion: .prefixedEnableDisable, help: typeOnlyHelp)
   var typeOnly: Bool
 
+  @Option(parsing: .upToNextOption, help: argumentNameHelp)
+  var argumentName: [String]
+
   /// Runs the command
   func run() throws {
     let cachedSyntaxTree = CachedSyntaxTree()
@@ -48,7 +51,7 @@ final class InitializerCommand: ParsableCommand {
     let fileURL = URL(fileURLWithPath: path)
 
     let initializerStatements = try analyzer.analyze(fileURL: fileURL)
-    let outputArray = initializerStatements.map { outputString(from: $0) }
+    let outputArray = initializerStatements.filter(shouldReturnParameters).map { outputString(from: $0) }
 
     let output = outputArray.joined(separator: "\n")
     print(output)
@@ -76,9 +79,27 @@ final class InitializerCommand: ParsableCommand {
       return statement.parameters.map { "\($0.name),\($0.typeNames.joined(separator: ","))" }.joined(separator: " ")
     }
   }
+
+  /// Filters the provided `InitializerStatement` given the user provided `argumentName`
+  ///
+  /// - Returns: `true` if the provided initializer statement should be used as output, `false` otherwise
+  private func shouldReturnParameters(from statement: InitializerStatement) -> Bool {
+    guard !argumentName.isEmpty else {
+      return true
+    }
+
+    let parameterNames = statement.parameters.map { $0.name }
+
+    return parameterNames.sorted().elementsEqual(argumentName.sorted())
+  }
 }
 
 private var typeOnlyHelp = ArgumentHelp("The granularity of the output",
                                         discussion: """
                                         Outputs a list of the type names by default. If disabled it outputs the name of the parameter and the name of the type (e.g. 'foo,Int bar,String')
+                                        """)
+
+private var argumentNameHelp = ArgumentHelp("A list of arguments name to filter initializers for",
+                                        discussion: """
+                                        When this value is provided, the command will only return the initializers that contain the list of arguments provided and will filter out everything else
                                         """)
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -80,7 +80,7 @@ final class InitializerCommand: ParsableCommand {
     }
   }
 
-  /// Filters the provided `InitializerStatement` given the user provided `argumentName`
+  /// Filters the provided `InitializerStatement` given the user-provided `argumentName`
   ///
   /// - Returns: `true` if the provided initializer statement should be used as output, `false` otherwise
   private func shouldReturnParameters(from statement: InitializerStatement) -> Bool {
--- a/Sources/SwiftInspectorCommands/InitializerCommand.swift
+++ b/Sources/SwiftInspectorCommands/InitializerCommand.swift
@@ -42,7 +42,7 @@ final class InitializerCommand: ParsableCommand {
   var typeOnly: Bool
 
   @Option(parsing: .upToNextOption, help: argumentNameHelp)
-  var argumentName: [String]
+  var parameterName: [String]
 
   /// Runs the command
   func run() throws {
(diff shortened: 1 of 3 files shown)


Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -76,9 +79,27 @@ final class InitializerCommand: ParsableCommand {
       return statement.parameters.map { "\($0.name),\($0.typeNames.joined(separator: ","))" }.joined(separator: " ")
     }
   }
+
+  /// Filters the provided `InitializerStatement` given the user provided `argumentName`
```
dfed, line 83: extreme nit.
```suggestion
  /// Filters the provided `InitializerStatement` given the user-provided `argumentName`
```

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -124,7 +129,25 @@ public struct InitializerStatement: Equatable {
...
+    }
+
+    public init(name: String, typeNames: [String]) {
+      self.name = name
+      self.typeNames = typeNames
+    }
+
+    /// The most common use case is for a parameter to only contain one type
+    /// That's why we have this convenience initializer
+    public init(name: String, typeName: String) {
```
bachand, line 147: Do we use this one anywhere?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -124,7 +129,25 @@ public struct InitializerStatement: Equatable {
 
   public struct Parameter: Equatable {
     public let name: String
-    public let typeName: String
+    public let typeNames: [String]
+
+    /// The most common use case is for a parameter to only contain one type
+    /// That's why we have this computed property as a convenience
+    public var typeName: String? {
```
bachand, line 136: Do we use this anywhere?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
bachand: I guess what I'm getting at is that it's not clear we should have if we never us it, and I wasn't seeing it.

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -41,14 +41,17 @@ final class InitializerCommand: ParsableCommand {
   @Flag(name: .shortAndLong, default: true, inversion: .prefixedEnableDisable, help: typeOnlyHelp)
   var typeOnly: Bool
 
+  @Option(parsing: .upToNextOption, help: argumentNameHelp)
+  var argumentName: [String]
```
bachand, line 45: Why not call it `paramaterName`? Also should it be plural "names" given that it's an array?

Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
```diff
@@ -87,6 +87,33 @@ final class InitializerCommandSpec: QuickSpec {
...
+
+        beforeEach {
+          fileURL = try? Temporary.makeFile(content: """
+          final class Some {
+            init(some: String, someInt: Int) {}
+          }
+          """)
+        }
+
+        it("filters out the initializers that don't have the argument names") {
```
bachand, line 101: What happens if one of the parameter names is `_`?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
fdiaz: We are! All over the tests https://github.com/fdiaz/SwiftInspector/blob/f3e16254429cc08033b07157ac5274efe3b25d4e/Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift#L87

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
fdiaz: Yup! On tests https://github.com/fdiaz/SwiftInspector/blob/f3e16254429cc08033b07157ac5274efe3b25d4e/Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift#L75

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: So this can be used from the command line like `--argument-name Some` or with multiple parameters like `--argument-name Some Another`. You're right that parameter name is probably a better name, I'll update.

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
bachand: Huh I wonder if these should be in an extension for tests?

fdiaz: I still feel like this convenience makes sense. Most of the initializers have a single parameter.

Sources/SwiftInspectorCommands/InitializerCommand.swift
fdiaz: https://github.com/fdiaz/SwiftInspector/pull/22/commits/91dfffece62faf487cefe7623940f2b6188560bf

Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
fdiaz: I haven't consider argument labels or `_` at all. I added this to track it https://github.com/fdiaz/SwiftInspector/issues/24 since it's out of the scope of this particular PR.

Good call though!

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
bachand: I guess when you think about _writing_ code this makes a lot of sense 🤔

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -139,6 +139,43 @@ final class InitializerAnalyzerSpec: QuickSpec {
             ]
           }
         }
+
+        context("with a tuple") {
+          beforeEach {
+            let content = """
+            public final class FakeType {
+              init(someTyple: (String, Int)) {}
```
bachand, line 147: `someTyple` -> `someTuple`

Sources/SwiftInspectorCore/Tests/InitializerAnalyzerSpec.swift
```diff
@@ -139,6 +139,43 @@ final class InitializerAnalyzerSpec: QuickSpec {
...
+
+          it("returns all the types in the array") {
+            let result = try? sut.analyze(fileURL: fileURL)
+            expect(result?.first?.parameters) == [
+              InitializerStatement.Parameter(name: "someTyple", typeNames: ["String", "Int"]),
+            ]
+          }
+        }
+
+        context("with a closure") {
```
bachand, line 161: We can delete this, right, since it's a tuple as well?

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -42,7 +42,7 @@ final class InitializerCommand: ParsableCommand {
   var typeOnly: Bool
 
   @Option(parsing: .upToNextOption, help: argumentNameHelp)
-  var argumentName: [String]
+  var parameterName: [String]
```
bachand, line 45: 👍

PR #26, opened 2020-04-04
Add Makefile to simplify setup

fdiaz:
Heavily inspired by [this article](https://nshipster.com/homebrew/#creating-a-makefile) I decided to simplify the installation and development of SwiftInspector.

We might want to put this in Homebrew at some point, but not today.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+prefix ?= /usr/local
+bindir ?= $(prefix)/bin
+
+.PHONY: develop
+develop:
+	swift package generate-xcodeproj
+	open SwiftInspector.xcodeproj 
+
+.PHONY: build
+build:
+	swift build -c release
+
+.PHONY: install
+install:
+	build
+	mv ".build/release/SwiftInspector" ".build/release/swift-inspector" 
+	install ".build/release/swift-inspector" "$(bindir)"
+
+.PHONY: uninstall
+uninstall:
+	rm -rf "$(bindir)/swift-inspector"
+
+.PHONY: clean
+clean:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,25 @@
+prefix ?= /usr/local
+bindir ?= $(prefix)/bin
+
+.PHONY: develop
+develop:
+	swift package generate-xcodeproj
+	open SwiftInspector.xcodeproj 
+
+.PHONY: build
```
bachand, line 9: What do you think of adding a debug target to build in debug?

README.md
```diff
@@ -12,19 +12,28 @@ This project is currently under development and can have breaking API changes.
 
 ---
 
-## Generating an Xcode Project
+## Requirements
 
-To see this project locally in Xcode you can run the following command on the root of this project:
-`swift package generate-xcodeproj`
+- Swift 5.1
```
bachand, line 17: Could be also to list the version of Xcode you need to have.

fdiaz: It's not trully _required_ to use Xcode. You could develop with something like VS Code, but it would still require Swift 5.1., so I'd prefer to keep this constraint to the actual requirement of the codebase.

Makefile
fdiaz: I though about adding one for building in debug, running the CLI and running tests but given those commands are

`swift build`
`swift test`
`swift run`

adding a rape to make it look like:

`make debug`
`make test`
`make run`

didn't provide a lot of value IMO, which is why I didn't add these.

PR #33, opened 2020-04-06
Update swift 5.2

fdiaz:
Updates `ArgumentParser` and `SwiftSyntax` to be able to be used in Swift 5.2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,8 @@ jobs:
     steps:
     # Checks-out the repo. More at: https://github.com/actions/checkout
     - uses: actions/checkout@v2
+    - name: Select Xcode version
+      run: sudo xcode-select -switch /Applications/Xcode_11.4.app
     - name: Build project
       run: swift build
     - name: Run tests and gather code coverage
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.2
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
--- a/Package.swift
+++ b/Package.swift
@@ -14,7 +14,7 @@ let package = Package(
     .library(name: "SwiftInspectorCore", targets: ["SwiftInspectorCore"])
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.1")),
+    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.4")),
     .package(url: "https://github.com/apple/swift-syntax.git", .exact("0.50100.0")),
     .package(url: "https://github.com/Quick/Nimble.git", .upToNextMajor(from: "8.0.1")),
     .package(url: "https://github.com/Quick/Quick.git", .upToNextMajor(from: "2.0.0")),
--- a/Package.swift
+++ b/Package.swift
@@ -15,7 +15,7 @@ let package = Package(
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.4")),
-    .package(url: "https://github.com/apple/swift-syntax.git", .exact("0.50100.0")),
+    .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", .exact("0.50200.0")),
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -8,6 +8,8 @@ jobs:
     steps:
     # Checks-out the repo. More at: https://github.com/actions/checkout
     - uses: actions/checkout@v2
+    - name: Select Xcode version
```
bachand, line 11: Maybe we can comment that this can be removed eventually?

Package.swift
```diff
@@ -14,8 +14,8 @@ let package = Package(
     .library(name: "SwiftInspectorCore", targets: ["SwiftInspectorCore"])
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.1")),
-    .package(url: "https://github.com/apple/swift-syntax.git", .exact("0.50100.0")),
+    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.4")),
+    .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", .exact("0.50200.0")),
```
fdiaz, line 18: It's now required that you pass the name for this. CI was failing if we didn't.

Sources/SwiftInspectorCore/ImportsAnalyzer.swift
```diff
@@ -97,7 +97,7 @@ public final class ImportsAnalyzer: Analyzer {
     var submoduleIdentifier: String = ""
 
     for child in syntaxNode.children {
-      guard let accessPath = child as? AccessPathSyntax else {
+      guard let accessPath = child.as(AccessPathSyntax.self) else {
```
bachand, line 100: Can you explain this to me?

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -36,11 +36,11 @@ public final class InitializerAnalyzer: Analyzer {
     var statements: [InitializerStatement] = []
     let syntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
 
-    var reader = InitializerSyntaxReader(shouldVisitIdentifier: shouldVisit) { [unowned self] node in
+    let reader = InitializerSyntaxReader(shouldVisitIdentifier: shouldVisit) { [unowned self] node in
       let statement = self.initializerStatement(from: node)
       statements.append(statement)
     }
-    syntax.walk(&reader)
+    reader.walk(syntax)
```
bachand, line 43: This syntax makes more sense anyway

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
```diff
@@ -192,19 +192,19 @@ private final class InitializerSyntaxReader: SyntaxVisitor {
     self.onNodeVisit = onNodeVisit
   }
 
-  func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+  override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
```
bachand, line 195: So it's a subclass and not a protocol?

bachand:
Any other interesting learnings from the release notes?

.github/workflows/ci.yml
fdiaz: I think we want to keep being explicit about what Xcode version we're using given the Swift requirement. I hate that we need to do this so hacky, but I didn't find a better way to do it with GitHub actions.

Sources/SwiftInspectorCore/ImportsAnalyzer.swift
fdiaz: @bachand how Swiftsyntax structures nodes moved from being protocol based to struct based. You can read more about [this here](https://github.com/apple/swift-syntax/blob/master/Changelog.md#modelling-syntax-nodes-as-structs)

Sources/SwiftInspectorCore/InitializerAnalyzer.swift
fdiaz: Yes, this was changed with the new version https://github.com/apple/swift-syntax/blob/master/Changelog.md#swift-52

bachand: Cool release notes are fairly helpful

PR #35, opened 2020-07-20
Add a way to return the conforming type when searching for type conformance

fdiaz:
**This PR is easier to review commit-by-commit**

The type-conformance command didn't return the information of the conforming type, only if the conformance existed in the file or not.

This PR introduces a way to output the names of the conforming types too.

That is, when a file contains:

```swift
protocol Some {}

final class Foo: Some { }
```

We now also return `Foo` in the output.

### DIFF
--- a/Sources/SwiftInspectorCore/Tests/TypeConformanceAnalyzerSpec.swift
+++ b/Sources/SwiftInspectorCore/Tests/TypeConformanceAnalyzerSpec.swift
@@ -40,7 +40,7 @@ final class TypeConformanceAnalyzerSpec: QuickSpec {
     }
     
     describe("analyze(fileURL:)") {
-      var result: TypeConformance!
+      var result: TypeConformance?
       
       context("when a type conforms to a protocol") {
         context("with only one conformance") {
--- a/Sources/SwiftInspectorCore/Tests/TypeConformanceAnalyzerSpec.swift
+++ b/Sources/SwiftInspectorCore/Tests/TypeConformanceAnalyzerSpec.swift
@@ -76,7 +76,7 @@ final class TypeConformanceAnalyzerSpec: QuickSpec {
               """
 
               self.fileURL = try? Temporary.makeFile(content: content)
-              let sut = TypeConformanceAnalyzer(typeName: "Bar")
+              let sut = TypeConformanceAnalyzer(typeName: "Foo")
               result = try? sut.analyze(fileURL: self.fileURL)
             }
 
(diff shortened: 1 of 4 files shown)


Sources/SwiftInspectorCore/Tests/TypeConformanceAnalyzerSpec.swift
```diff
@@ -40,26 +40,28 @@ final class TypeConformanceAnalyzerSpec: QuickSpec {
     }
 
     describe("analyze(fileURL:)") {
+      var result: TypeConformance!
```
dfed, line 43: do you need the force unwrap? I looks like you use `?` when inspecting this value below

Sources/SwiftInspectorCore/TypeConformanceAnalyzer.swift
```diff
@@ -56,6 +62,24 @@ public final class TypeConformanceAnalyzer: Analyzer {
...
   }
+
+  private func findConformingType(of node: SyntaxProtocol?) -> String? {
+    guard let originalNode = node else { return nil }
+
+    guard let parent = originalNode.parent else { return nil }
+
+    // A conforming type can only be a class, struct or enum
+    // See: https://docs.swift.org/swift-book/LanguageGuide/Protocols.html
+    if let classSyntax = parent.as(ClassDeclSyntax.self) {
```
dfed, line 73: asking to learn: where does the `func as` method come from?

dfed:
It would be nice update the README to include this feature before merging, but that can be done in a follow-up.

Sources/SwiftInspectorCore/Tests/TypeConformanceAnalyzerSpec.swift
fdiaz: Yup! Good idea

Sources/SwiftInspectorCore/TypeConformanceAnalyzer.swift
fdiaz: @dfed this is from SwiftSyntax, it was just introduced in the Swift 5.2 version, [see here for more](https://github.com/apple/swift-syntax/blob/master/Changelog.md#swift-52)

fdiaz:
> It would be nice update the README to include this feature before merging, but that can be done in a follow-up.

We actually don't have a great documentation of these commands. These is something @bachand has asked for a couple of times. I created https://github.com/fdiaz/SwiftInspector/issues/37 to track this.

PR #43, opened 2020-08-17
add property analyzer

thedrick:
## Changes

This adds a new property analyzer and command. As with the `TypeCommand` / `TypeAnalyzer` this has more information in the analyzer than it does in the command.

Here's the command
```
swift-inspector properties --help
OVERVIEW: Finds property information for the provided type

USAGE: SwiftInspector properties --name <name> --path <path>

OPTIONS:
  --name <name>           The name of the type to find property information on 
        This may be a enum, class, struct, or protocol.
  --path <path>           The absolute path of the file to inspect 
```

This command will find property information for a provided type including:
- Type name
- Each property includes:
  - Property name
  - Access (public, private, internal, fileprivate)
  - Scope (instance, scope)
  - Comment (Analyzer only)
  - Property type (Analyzer only)

## Please review
@fdiaz

### DIFF
--- a/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
+++ b/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
@@ -1,0 +1,76 @@
+// Created by Tyler Hedrick on 8/17/20.
+//
+// Copyright (c) 2020 Tyler Hedrick
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import ArgumentParser
+import Foundation
+import SwiftInspectorCore
+
+final class PropertyAnalyzerCommand: ParsableCommand {
+  static var configuration = CommandConfiguration(
+    commandName: "properties",
+    abstract: "Finds property information for the provided type"
+  )
+
+  @Option(help: nameArgumentHelp)
+  var name: String
+
+  @Option(help: "The absolute path of the file to inspect")
+  var path: String
+
+  /// Runs the command
+  func run() throws {
+    let cachedSyntaxTree = CachedSyntaxTree()
+
+    let analyzer = PropertyAnalyzer(typeName: name, cachedSyntaxTree: cachedSyntaxTree)
+    let fileURL = URL(fileURLWithPath: path)
+    if let propertyInformation = try analyzer.analyze(fileURL: fileURL) {
+      print(outputString(from: propertyInformation))
+    }
+  }
+
+  /// Validates if the arguments of this command are valid
+  func validate() throws {
+    guard !name.isEmpty else {
+      throw InspectorError.emptyArgument(argumentName: "--name")
+    }
+    guard !path.isEmpty else {
+      throw InspectorError.emptyArgument(argumentName: "--path")
+    }
+    let pathURL = URL(fileURLWithPath: path)
+    guard FileManager.default.isSwiftFile(at: pathURL) else {
+      throw InspectorError.invalidArgument(argumentName: "--path", value: path)
+    }
+  }
+
+  private func outputString(from statement: TypeWithPropInfo) -> String {
+    let propString = statement.properties.map { propInfo in
+      "(\(propInfo.name)|\(propInfo.access)|\(propInfo.scope))"
+    }.joined(separator: ",")
+    return "\(statement.name):\(propString)"
+  }
+}
+
+private let nameArgumentHelp = ArgumentHelp(
+  "The name of the type to find property information on",
+  discussion: "This may be a enum, class, struct, or protocol.")
--- a/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
+++ b/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
@@ -65,7 +65,7 @@ final class PropertyAnalyzerCommand: ParsableCommand {
 
   private func outputString(from statement: TypeProperties) -> String {
     statement.properties.map { propInfo in
-      "\(propInfo.name),\(propInfo.modifiers)"
+      "\(statement.name),\(propInfo.name),\(propInfo.modifiers)"
     }.joined(separator: "\n")
   }
 }
@@ -73,3 +73,24 @@ final class PropertyAnalyzerCommand: ParsableCommand {
 private let nameArgumentHelp = ArgumentHelp(
   "The name of the type to find property information on",
   discussion: "This may be a enum, class, struct, or protocol.")
+
+extension TypeProperties.Modifier: CustomStringConvertible, CustomDebugStringConvertible {
+  public var description: String {
+    var outputValues: [String] = []
+    // Order is important here! Access control modifiers should always go before scope modifiers
+    // Access control modifiers
+    if contains(.public) { outputValues.append("public") }
+    if contains(.private) { outputValues.append("private") }
+    if contains(.fileprivate) { outputValues.append("fileprivate") }
+    if contains(.internal) { outputValues.append("internal") }
+    if contains(.privateSet) { outputValues.append("private(set)") }
+    if contains(.internalSet) { outputValues.append("internal(set)") }
+    if contains(.publicSet) { outputValues.append("public(set)") }
+    // Scope modifiers
+    if contains(.static) { outputValues.append("static") }
+    if contains(.instance) { outputValues.append("instance") }
+    return outputValues.joined(separator: ",")
+  }
+
+  public var debugDescription: String { description }
+}
--- a/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
+++ b/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
@@ -65,7 +65,7 @@ final class PropertyAnalyzerCommand: ParsableCommand {
 
   private func outputString(from statement: TypeProperties) -> String {
     statement.properties.map { propInfo in
-      "\(propInfo.name),\(propInfo.modifier)"
+      "\(propInfo.name),\(propInfo.modifiers)"
     }.joined(separator: "\n")
   }
 }
--- a/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
+++ b/Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
@@ -63,11 +63,10 @@ final class PropertyAnalyzerCommand: ParsableCommand {
     }
   }
 
-  private func outputString(from statement: TypeWithPropInfo) -> String {
-    let propString = statement.properties.map { propInfo in
-      "(\(propInfo.name)|\(propInfo.access)|\(propInfo.scope))"
-    }.joined(separator: ",")
-    return "\(statement.name):\(propString)"
+  private func outputString(from statement: TypeProperties) -> String {
+    statement.properties.map { propInfo in
+      "\(propInfo.name),\(propInfo.modifier)"
+    }.joined(separator: "\n")
   }
 }
 
--- a/Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
@@ -1,0 +1,106 @@
+// Created by Tyler Hedrick on 8/17/20.
+//
+// Copyright (c) 2020 Tyler Hedrick
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Nimble
+import Quick
+import Foundation
+
+@testable import SwiftInspectorCore
+
+final class PropertyAnalyzerCommandSpec: QuickSpec {
+  override func spec() {
+    describe("run") {
+
+      context("with no arguments") {
+        it("fails") {
+          let result = try? TestTask.run(withArguments: ["properties"])
+          expect(result?.didFail) == true
+        }
+      }
+
+      context("when path is invalid") {
+        it("fails when empty") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", ""])
+          expect(result?.didFail) == true
+        }
+
+        it("fails when not provided") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path"])
+          expect(result?.didFail) == true
+        }
+
+        it("fails when it doesn't exist") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", "/fake/path"])
+          expect(result?.didFail) == true
+        }
+
+        it("fails when name is present but the file is invalid") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", "/fake/path", "--name", "SomeType"])
+          expect(result?.didFail) == true
+        }
+      }
+
+      context("when name is passed") {
+        var fileURL: URL!
+
+        beforeEach {
+          fileURL = try? Temporary.makeFile(content: """
+          final class Some {
+            public var thing: String = "Hello, World"
+          }
+          """)
+        }
+
+        afterEach {
+          try? Temporary.removeItem(at: fileURL)
+        }
+
+        it("fails when name is empty") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name", ""])
+          expect(result?.didFail) == true
+        }
+
+        it("fails when name is not provided") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name"])
+          expect(result?.didFail) == true
+        }
+
+        it("fails when path is empty") {
+          let result = try? TestTask.run(withArguments: ["properties", "--name", "Some"])
+          expect(result?.didFail) == true
+        }
+
+        it("succeeds") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name", "Some"])
+          expect(result?.didSucceed) == true
+        }
+
+        it("returns the type name with property information including the name, access, and scope") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name", "Some"])
+          expect(result?.outputMessage).to(contain("Some:(thing|public|instance)"))
+        }
+      }
+    }
+  }
+}
--- a/Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
@@ -98,7 +98,7 @@ final class PropertyAnalyzerCommandSpec: QuickSpec {
 
         it("returns the type name with property information including the name, access, and scope") {
           let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name", "Some"])
-          expect(result?.outputMessage).to(contain("Some:(thing|public|instance)"))
+          expect(result?.outputMessage).to(contain("thing,public,instance"))
         }
       }
     }
--- a/Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
+++ b/Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
@@ -98,7 +98,7 @@ final class PropertyAnalyzerCommandSpec: QuickSpec {
 
         it("returns the type name with property information including the name, access, and scope") {
           let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name", "Some"])
-          expect(result?.outputMessage).to(contain("thing,public,instance"))
+          expect(result?.outputMessage).to(contain("Some,thing,public,instance"))
         }
       }
     }
--- a/Sources/SwiftInspectorCore/PropertyAnalyzer.swift
+++ b/Sources/SwiftInspectorCore/PropertyAnalyzer.swift
@@ -143,7 +143,7 @@ private final class PropertySyntaxVisitor: SyntaxVisitor {
             name: identifier.identifier.text,
             typeAnnotation: simpleTypeIdentifier.name.text,
             comment: comment(from: leadingTrivia),
-            modifier: modifier))
+            modifiers: modifier))
           return
         }
         // no type annotation was found. In this case it's much more complicated to get
@@ -156,7 +156,7 @@ private final class PropertySyntaxVisitor: SyntaxVisitor {
           name: identifier.identifier.text,
           typeAnnotation: nil,
           comment: comment(from: leadingTrivia),
-          modifier: modifier))
+          modifiers: modifier))
       }
     }
 
@@ -206,7 +206,10 @@ private final class PropertySyntaxVisitor: SyntaxVisitor {
     }
 
     // If there are no explicit modifiers, this is an internal property
-    if modifier.isEmpty {
+    if !modifier.contains(.public) &&
+      !modifier.contains(.fileprivate) &&
+      !modifier.contains(.private)
+    {
       modifier = modifier.union(.internal)
     }
 
@@ -260,7 +263,7 @@ public struct TypeProperties: Hashable {
     /// Any comments associated with the property
     public let comment: String
     /// Modifier set for this type
-    public let modifier: Modifier
+    public let modifiers: Modifier
   }
 
   /// The name of the type.
--- a/Sources/SwiftInspectorCore/PropertyAnalyzer.swift
+++ b/Sources/SwiftInspectorCore/PropertyAnalyzer.swift
@@ -23,7 +23,7 @@
 import Foundation
 import SwiftSyntax
 
-// MARK: - TypeLocationAnalyzer
+// MARK: - PropertyAnalyzer
 
 public final class PropertyAnalyzer: Analyzer {
 
@@ -34,14 +34,14 @@ public final class PropertyAnalyzer: Analyzer {
     self.cachedSyntaxTree = cachedSyntaxTree
   }
 
-  /// Finds the location(s) of the specified type name in the Swift file
+  /// Finds property information for the provided type in the Swift file
   /// - Parameter fileURL: The fileURL where the Swift file is located
-  public func analyze(fileURL: URL) throws -> TypeWithPropInfo? {
+  public func analyze(fileURL: URL) throws -> TypeProperties? {
     let syntax: SourceFileSyntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
-    var result: TypeWithPropInfo?
-    let visitor = TypeSyntaxVisitor() { [unowned self] typeWithPropInfo in
-      guard self.typeName == typeWithPropInfo.name else { return }
-      result = try? typeWithPropInfo.merge(with: result)
+    var result: TypeProperties?
+    let visitor = TypeSyntaxVisitor(typeName: typeName) { [unowned self] typeProperties in
+      guard self.typeName == typeProperties.name else { return }
+      result = try? typeProperties.merge(with: result)
     }
     visitor.walk(syntax)
     return result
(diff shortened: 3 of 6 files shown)


Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+    let pathURL = URL(fileURLWithPath: path)
+    guard FileManager.default.isSwiftFile(at: pathURL) else {
+      throw InspectorError.invalidArgument(argumentName: "--path", value: path)
+    }
+  }
+
+  private func outputString(from statement: TypeWithPropInfo) -> String {
+    let propString = statement.properties.map { propInfo in
+      "(\(propInfo.name)|\(propInfo.access)|\(propInfo.scope))"
```
fdiaz, line 68: Should we use spaces or commas to separate these and then new lines for each property instead? this would be the first command that uses`|` to separate fields

Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        it("succeeds") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name", "Some"])
+          expect(result?.didSucceed) == true
+        }
+
+        it("returns the type name with property information including the name, access, and scope") {
+          let result = try? TestTask.run(withArguments: ["properties", "--path", fileURL.path, "--name", "Some"])
+          expect(result?.outputMessage).to(contain("Some:(thing|public|instance)"))
```
fdiaz, line 101: I'd suggest outputting something like:

`Some thing public instance` instead. This is a lot easier to parse as output and to provide as arguments for another tool.

I really need to come up with better documentation for these things.

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+public final class PropertyAnalyzer: Analyzer {
+
+  /// - Parameter typeName: The name of the type to get property information for
+  /// - Parameter cachedSyntaxTree: The cached syntax tree to return the AST tree from
+  public init(typeName: String, cachedSyntaxTree: CachedSyntaxTree = .init()) {
+    self.typeName = typeName
+    self.cachedSyntaxTree = cachedSyntaxTree
+  }
+
+  /// Finds the location(s) of the specified type name in the Swift file
```
fdiaz, line 37: I don't think this is correct?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+      default: return nil
+      }
+    }.joined(separator: "\n")
+  }
+}
+
+// MARK: - TypeWithPropInfo
+
+/// Information about a located type. Indices start with 0.
+public struct TypeWithPropInfo: Hashable {
```
fdiaz, line 184: `TypeProperties` instead?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+
+  public enum Access: String {
```
fdiaz, line 186: nit: Let's name this AccessControl

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+    case `private`
+    case `fileprivate`
+  }
+
+  public enum Scope: String {
+    case instance
+    case `static`
+  }
+
+  public struct PropInfo: Hashable {
```
fdiaz, line 198: `PropertyInfo` https://swift.org/documentation/api-design-guidelines/#promote-clear-usage ?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+    public let comment: String
+    /// Access control of this property
+    public let access: Access
+    /// Scope of this property
+    public let scope: Scope
+  }
+
+  /// The name of the type.
+  public let name: String
+  /// The properites on this type
```
fdiaz, line 213: typo: properties

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+      case .docBlockComment(let str): return str
+      default: return nil
+      }
+    }.joined(separator: "\n")
+  }
+}
+
+// MARK: - TypeWithPropInfo
+
+/// Information about a located type. Indices start with 0.
```
fdiaz, line 183: I don't think this comment is correct?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+// MARK: - TypeLocationSyntaxVisitor
+
+private final class TypeSyntaxVisitor: SyntaxVisitor {
+
+  init(onNodeVisit: @escaping (_ info: TypeWithPropInfo) -> Void) {
+    self.onNodeVisit = onNodeVisit
+  }
+
+  override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+    processNode(node, withName: node.identifier.text, members: node.members.members)
```
fdiaz, line 65: Should we validate in here the name instead that after doing all the processing? That will likely improve the performance of this tool (since we won't be processing every type, only the types we care about)

I'mm thinking we should pass the type name to the `TypeSyntaxVisitor` and if the name is not the one expected on these methods we can do a guard and visit children immediately, and only process the node when the type name is the one we care about.

```swift
guard node.identifier.text == typeName else { return .visitChildren }
```

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+    return .visitChildren
+  }
+
+  override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+    processNode(node, withName: node.identifier.text, members: node.members.members)
+    return .visitChildren
+  }
+
+  override func visit(_ node: ExtensionDeclSyntax) -> SyntaxVisitorContinueKind {
+    if let typeIdentifier = node.extendedType.as(SimpleTypeIdentifierSyntax.self) {
```
fdiaz, line 85: Can you explain why we need this casting?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+  override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
+    var leadingTrivia: Trivia? = node.leadingTrivia
+    var access: TypeWithPropInfo.Access = .internal
+    var scope: TypeWithPropInfo.Scope = .instance
+
+    node.modifiers?.forEach { modifier in
+      // leading trivia is on the first modifier, or the node itself if no modifiers are present
+      leadingTrivia = leadingTrivia ?? modifier.leadingTrivia
+      // This attempts to get any access information from the node's modifiers
+      if let modifierAccess = TypeWithPropInfo.Access(rawValue: modifier.name.text) {
```
fdiaz, line 123: Should we be looking into DeclModifier syntax instead? I think this could break if we have an attribute like `public private(set)` since we'd be first assigning the modifier to public, but later re-assigning to private.

I think it'd be better to look at the tokens of each declaration modifier and look for the public, internal, etc keywords, then we could validate that we only have 1 token (so we make sure it's only public and not something like public(set)`?

I've found that assuming ordering when using the AST usually ends up biting me back. I think using the token names tends to be more robust and scales better.

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+        access = modifierAccess
+      }
+      // this attempts to get any scope information from the node's modifiers
+      if let modifierScope = TypeWithPropInfo.Scope(rawValue: modifier.name.text) {
```
fdiaz, line 127: same here, probably better to look for the static keyword in the tokens of the Decl modifier

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+        scope = modifierScope
+      }
+    }
+
+    // in practice this `bindings` array only has 1 element
```
fdiaz, line 132: i don't think this is always true, I think we can have something like:

```swift
struct Foo {
  let some: String, another: Int
}
```

And this will contain 2 bindings IIUC

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+            name: identifier.identifier.text,
+            typeAnnotation: simpleTypeIdentifier.name.text,
+            comment: comment(from: leadingTrivia),
+            access: access,
+            scope: scope))
+          return
+        }
+        // no type annotation was found. In this case it's much more complicated to get
+        // the type information for a variable.
+        // TODO: Include logic to get types of any variable declaration
```
fdiaz, line 149: What's an example of this use case?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+          scope: scope))
+      }
+    }
+
+    return .skipChildren
+  }
+
+  // MARK: Private
+
+  private let parentName: String
```
fdiaz, line 164: nit: `typeName`?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  /// Finds the location(s) of the specified type name in the Swift file
+  /// - Parameter fileURL: The fileURL where the Swift file is located
+  public func analyze(fileURL: URL) throws -> TypeWithPropInfo? {
+    let syntax: SourceFileSyntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
+    var result: TypeWithPropInfo?
+    let visitor = TypeSyntaxVisitor() { [unowned self] typeWithPropInfo in
+      guard self.typeName == typeWithPropInfo.name else { return }
+      result = try? typeWithPropInfo.merge(with: result)
```
fdiaz, line 44: Should we rethrow this error instead of ignoring it?

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+              name: "thing",
+              typeAnnotation: "String",
+              comment: "",
+              access: .public,
+              scope: .static)
+          ]
+        }
+      }
+
+      context("when there is a private property") {
```
fdiaz, line 301: Let's also add a test for wwhen these are the other way: `static private` instead of `private static`

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+          expect(result?.properties) == [
+            TypeWithPropInfo.PropInfo(
+              name: "thing",
+              typeAnnotation: "String",
+              comment: "/** The thing */",
+              access: .public,
+              scope: .instance)
+          ]
+        }
+      }
```
fdiaz, line 464: Let's add a test for the `public private(set)` case too

bachand:
I will take a look at this tomorrow. I love that we are investing more in Swift Inspector!

Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
thedrick: Yeah, i wasn't sure what to do here exactly, im good with changing it to what you said 👍

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
thedrick: lol copy pasta 🙈

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
thedrick: I wish I could! I don't really understand SwiftSyntax, I've been using this tool to help me: https://swift-ast-explorer.kishikawakatsumi.com/. The `ExtensionDeclSyntax` doesn't have information on the type that it extends directly, at least not from what I could tell. If you have more information on how to do this I'd love to improve it

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
thedrick: At this point we are already inside of the `VariableDeclSyntax` node. Maybe I just don't understand how the node API works, but wouldn't your example result in 2 separate nodes being parsed instead of 1?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
thedrick: I can include this in the comments, but this is an example where we can easily find the type:
```swift
public static let thing: String = "Hello"
```
whereas this doesn't have the type information:
```swift
public static let thing = "Hello"
```
I don't really know how to handle the latter case here, so I've just chosen to pass over it for the time being. What are your thoughts?

thedrick:
Thanks for the thorough review @fdiaz! I'll update this tomorrow

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
fdiaz: I don't think so. That'd be one variable declaration IIUC with two bindings but I haven't validated this. maybe we can add a test for this use case?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
fdiaz: I think it makes sense to not pass the type here, This makes sense! Let's add this as a comment 🙏

Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
```diff
@@ -0,0 +1,106 @@
...
+
+        beforeEach {
+          fileURL = try? Temporary.makeFile(content: """
+          final class Some {
+            public var thing: String = "Hello, World"
+          }
+          """)
+        }
+
+        afterEach {
```
bachand, line 75: 🏆

Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
bachand: So let's say that you ran this command like `./SwiftInspector properties -name Some -path /whatever.swift` and the type looked like this:

```swift
class Some {
  private let hello: Int
  public let goodbye: String
}
```

I think output like the following would look be good.

```
hello, private, instance
goodbye, public, instance
```

I don't think we need to regurgitate the type name since you passed it explicitly into the command.

(Apologies if I don't _fully_ understand the output but I _think_ this close and should give an idea of where my head is at 👍 )

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import SwiftSyntax
+
+// MARK: - TypeLocationAnalyzer
```
bachand, line 26:
```suggestion
// MARK: - PropertyAnalyzer
```

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
bachand: I remember this :)

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+    self.cachedSyntaxTree = cachedSyntaxTree
+  }
+
+  /// Finds the location(s) of the specified type name in the Swift file
+  /// - Parameter fileURL: The fileURL where the Swift file is located
+  public func analyze(fileURL: URL) throws -> TypeWithPropInfo? {
+    let syntax: SourceFileSyntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
+    var result: TypeWithPropInfo?
+    let visitor = TypeSyntaxVisitor() { [unowned self] typeWithPropInfo in
+      guard self.typeName == typeWithPropInfo.name else { return }
```
bachand, line 43: When would the name be optional?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+    return result
+  }
+
+  // MARK: Private
+
+  private let cachedSyntaxTree: CachedSyntaxTree
+  private let typeName: String
+}
+
+// MARK: - TypeLocationSyntaxVisitor
```
bachand, line 56:
```suggestion
// MARK: - TypeSyntaxVisitor
```

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,248 @@
...
+      result.append(info)
+    }
+    propertyVisitor.walk(node)
+    onNodeVisit(.init(name: name, properties: result))
+  }
+}
+
+// MARK: - PropertySyntaxVisitor
+
+private final class PropertySyntaxVisitor: SyntaxVisitor {
```
bachand, line 107: I like the modularity here.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    afterEach {
+      guard let fileURL = fileURL else {
+        return
+      }
+      try? Temporary.removeItem(at: fileURL)
+    }
+
+    describe("TypeWithPropInfo.merge(other:)") {
```
bachand, line 47: In this style of testing I think it would be more idiomatic to make a `TypeWithPropInfoSpec` top-level class for testing this related type, as opposed to nesting it within this spec class. But if we do want to keep it here I think it would be good to have two describe invocations here: one for the type name, and one for the method we're testing.

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
bachand: I guess it's when we can't find the type were looking for?

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+        it("returns nil if the type name is not present") {
+          let sut = PropertyAnalyzer(typeName: "AnotherType")
+          let result = try? sut.analyze(fileURL: fileURL)
+          expect(result).to(beNil())
+        }
+
+        it("detects the properties") {
+          let result = try? sut.analyze(fileURL: fileURL)
+          expect(result?.properties) == [
+            TypeWithPropInfo.PropInfo(
```
bachand, line 132: 🔥

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+              name: "thing",
+              typeAnnotation: "String",
+              comment: "",
+              access: .internal,
+              scope: .instance)
+          ]
+        }
+      }
+
+      context("when there is a property extension") {
```
bachand, line 226: 👍

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+              name: "thing",
+              typeAnnotation: "String",
+              comment: "",
+              access: .internal,
+              scope: .instance)
+          ]
+        }
+      }
+
+      context("when there are multiple properites") {
```
bachand, line 256:
```suggestion
      context("when there are multiple properties") {
```

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+              typeAnnotation: "String",
+              comment: "",
+              access: .internal,
+              scope: .instance)
+          ]
+        }
+      }
+
+      context("when there is a property extension") {
+        beforeEach {
```
bachand, line 227: Could be cool to have a test for multiple extensions.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+        it("returns type info with empty property list") {
+          let result = try? sut.analyze(fileURL: fileURL)
+          expect(result?.properties).to(beEmpty())
+        }
+      }
+
+      context("when there is a property") {
+        beforeEach {
+          let content = """
+          public final class FakeType {
```
bachand, line 111: What would happen if I had this (valid) Swift code?

```swift
class FakeType {

  let hello: Int = 1

  class FakeType {
    let goodbye: String = ""
  }
}
```

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
bachand, line 107: Full transparency: I am going to defer to @fdiaz on reviewing this class. @fdiaz happy to dig in to help with specific questions/discussions.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
bachand: Here's how I handled that case for type location but it's a bit different here https://github.com/fdiaz/SwiftInspector/blob/a173d02d773af433cc9e36c5797f58a5f7735445/Sources/SwiftInspectorCore/Tests/TypeLocationAnalyzerSpec.swift#L295-L317

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+              name: "thing",
+              typeAnnotation: nil,
+              comment: "",
+              access: .private,
+              scope: .static)
+          ]
+        }
+      }
+
+      context("when there is a line comment") {
```
bachand, line 370: I don't know the inner workings of SwiftSyntax well enough but would we want to test a multi-line comment? @fdiaz you know better than me if that's getting into testing the internals of SwiftSyntax.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,468 @@
...
+              comment: "/** The thing */",
+              access: .public,
+              scope: .instance)
+          ]
+        }
+      }
+
+    }
+  }
+}
```
bachand, line 468: It could be nice to test that this works correctly when a property has an attribute.

```swift
class Foo {
  @objc let hello = "world"
}
```

bachand: And we may even want to include the attribute in our result.

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
thedrick: @fdiaz is there an easyish way to do that? This is inside of a block which is passed into another class and then called from a non-throwing override function. It gets really messy if we don't swallow this error

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
thedrick: yeah, exactly

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
thedrick: I can certainly test it, I just didn't think it was an interesting or useful test given the code

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
thedrick: hmm would it be ok to add that in a follow up? This feels like a use case that would be uncommon to need. I will add a test for it, but I don't want to add it to the output at this point.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
thedrick: @bachand I would argue that is very poor API design 😅 

I don't know, what do you think we should do here? I can handle the case just fine that you posted below where the nested class type is different from the outer class, but I don't know how to handle the nested class having the same name as the outer class. That just feels *super* rare to the point we don't need to cover it.

thedrick: @bachand yeah this is a great point... I don't currently handle this situation well - we will just merge all of the properties into one list. Do you have suggestions of how we should handle this? I could append the parent type name to the child type like:
```swift
FakeType.FakeType
```
to disambiguate. You would then need to pass in `FakeType.FakeType` in order to get property information from a nested type.

bachand: The example I gave is pretty contrived.

One that is maybe (a bit) more realistic could look like this.

```swift
class MyClass {

  let hello: Int = 1

  class NestedType {
    let goodbye: String = ""
  }
}

class YourClass {

  let foo: Int = 1

  class NestedType {
    let bar: String = ""
  }
}
```

Ideally we could look up information on `MyClass.NestedType` and `YourClass.NestedType` independently.

Appending the parent type name definitely seems like the Right™ way to do this. I didn't think of that!

thedrick: I just tried doing this and I'm not even sure how to accomplish what I suggested. Very open to suggestions on how to handle this!

thedrick:
@fdiaz this is ready for another pass. I've address all of the comments, the only outstanding issue is the nested type one that Michael pointed out. I wonder if that could be addressed in a follow-up?

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
bachand: Haha, yeah this sounds hard to me. I think you'd have to keep track of the parent type(s) as you traverse the nodes.

I can live with this limitation for now. It could be nice to have a test to show how we handle this case. Then we could add a comment to that test linking to an issue where we describe the ideal behavior.

cc: @fdiaz

Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
fdiaz: I think returning the type is useful when parsing the result. That way you don't need to inspect both what you're passing and what the output is giving you. I've found it useful at least

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
fdiaz: I think this is fine for now. Let's keep it as is and if we see problems in the real world or a use case where we need to handle this we can build on top of this.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
fdiaz: I agree, let's just add it as a follow-up if needed.

Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
bachand: I think it probably depends on whether you are invoking this command from another script (e.g. Ruby) or running it manually on the command line.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
bachand: That's fine by me.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -267,9 +216,46 @@ final class PropertyAnalyzerSpec: QuickSpec {
...
+              name: "foo",
+              typeAnnotation: "Int",
+              comment: "",
+              modifier: [.internal, .instance])
+          ]
+          expect(propSet) == expectedPropSet
+        }
+      }
+
+      context("when there are multiple properties on the same line") {
```
bachand, line 279: Nice

Sources/SwiftInspectorCommands/PropertyAnalyzerCommand.swift
thedrick: @fdiaz should I add the type information back into the output?

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -357,12 +384,33 @@ final class PropertyAnalyzerSpec: QuickSpec {
...
+            TypeProperties.PropertyData(
+              name: "thing",
+              typeAnnotation: nil,
+              comment: "",
+              modifier: [.private, .static])
+          ]
+        }
+      }
+
+      context("when there is a property attribute") {
```
bachand, line 440: This looks good for now. Comforting to know we handle the case even if we don't pull out the attribute itself yet.

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -381,12 +429,35 @@ final class PropertyAnalyzerSpec: QuickSpec {
...
               typeAnnotation: "String",
               comment: "// The thing",
-              access: .public,
-              scope: .instance)
+              modifier: [.public, .instance])
+          ]
+        }
+      }
+
+      context("when there is a multi-line comment") {
```
bachand, line 485: Thanks!

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -87,11 +87,55 @@ final class PropertyAnalyzerSpec: QuickSpec {
...
+        it("detects the type name") {
+          let result = try? sut.analyze(fileURL: fileURL)
+          expect(result?.name) == "FakeType"
+        }
+
+        /*
+         This is actually not the ideal result of this and is a limitation of the implementation
+         Ideally you would have to pass in `FakeType.FakeType` to get this nested type's property
+         information. For now we are accepting this limitation and have this test to showcase
+         what happens in this scenario.
```
bachand, line 118: Perfect 👌

bachand: I increasingly feel like unit tests are a great way to document edge cases and accepted limitations.

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+}
+
+extension TypeProperties.Modifier: CustomStringConvertible, CustomDebugStringConvertible {
+  public var description: String {
+    var outputValues: [String] = []
+    // Order is important here! Access control modifiers should always go before scope modifiers
+    // Access control modifiers
+    if contains(.public) { outputValues.append("public") }
+    if contains(.private) { outputValues.append("private") }
+    if contains(.privateSet) { outputValues.append("private(set)") }
```
fdiaz, line 314: Should setters go after `fileprivate` and `internal`?

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+    guard name == other.name else {
+      throw MergeError.invalidNames
+    }
+    return TypeProperties(
+      name: name,
+      properties: Array(Set(properties).union(Set(other.properties))))
+  }
+}
+
+extension TypeProperties.Modifier: CustomStringConvertible, CustomDebugStringConvertible {
```
fdiaz, line 307: Let's move this extension to the `Command` since this is related to how we output things to the standard output which the Analyzer shouldn't care about see: https://github.com/fdiaz/SwiftInspector/blob/main/CONTRIBUTING.md#swiftinspectorcommand

Sources/SwiftInspectorCore/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -0,0 +1,580 @@
...
+              comment: "",
+              modifiers: [.public, .instance]),
+            .init(
+              name: "foo",
+              typeAnnotation: "String",
+              comment: "",
+              modifiers: [.internal, .static])
+          ]
+          expect(propSet) == expectedPropSet
+        }
```
fdiaz, line 136: Agreed with Michael, it's great to have a test like this and document that it's not ideal but it's what a workaround we're ok with.

Sources/SwiftInspectorCore/PropertyAnalyzer.swift
```diff
@@ -0,0 +1,324 @@
...
+/// Information about a type as well as its associated properties
+public struct TypeProperties: Hashable {
+
+  public struct Modifier: Hashable, OptionSet {
+    public let rawValue: Int
+
+    public static let `internal` = Modifier(rawValue: 1 << 0)
+    public static let `public` = Modifier(rawValue: 1 << 1)
+    public static let `private` = Modifier(rawValue: 1 << 2)
+    public static let privateSet = Modifier(rawValue: 1 << 3)
```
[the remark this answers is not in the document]
thedrick: i didn't realize public(set) was a thing, but yes let's add those!

fdiaz:
Thank you!

PR #47, opened 2020-10-04
Update to use Swift 5.3

fdiaz:
This PR supersedes https://github.com/fdiaz/SwiftInspector/pull/46 by using the new release versions of all the dependencies which makes this change a much easier one.


dfed (approved):

Sources/SwiftInspectorCommands/InitializerCommand.swift
```diff
@@ -38,11 +38,11 @@ final class InitializerCommand: ParsableCommand {
...
   var name: String
 
-  @Flag(name: .shortAndLong, default: true, inversion: .prefixedEnableDisable, help: typeOnlyHelp)
-  var typeOnly: Bool
+  @Flag(name: .shortAndLong, inversion: .prefixedEnableDisable, help: typeOnlyHelp)
+  var typeOnly: Bool = true
 
   @Option(parsing: .upToNextOption, help: argumentNameHelp)
-  var parameterName: [String]
+  var parameterName: [String] = []
```
fdiaz, line 45: SwiftArgumentParser before defaulted arrays to an empty array, see https://github.com/apple/swift-argument-parser/blob/main/CHANGELOG.md#changes

settings.xcconfig
```diff
@@ -0,0 +1 @@
+CLANG_ENABLE_MODULES = YES;
```
fdiaz, line 1: See https://github.com/Quick/Quick/issues/707

Sources/SwiftInspectorCommands/StaticUsageCommand.swift
```diff
@@ -53,9 +53,6 @@ final class StaticUsageCommand: ParsableCommand {
 
   /// Validates if the arguments of this command are valid
   func validate() throws {
-    guard !statics.isEmpty else {
-      throw InspectorError.emptyArgument(argumentName: "--statics")
-    }
```
fdiaz, line 58: We don't need this validation since now this is already being done by SwiftArgumentParser https://github.com/apple/swift-argument-parser/blob/main/CHANGELOG.md#changes

I noticed this because we had a test for this use case and the code coverage dropped, since now it's actually impossible to trigger this path since it will be caught by SwiftArgumentParser before getting here.

Sources/SwiftInspectorCommands/TypeConformanceCommand.swift
```diff
@@ -53,9 +53,6 @@ final class TypeConformanceCommand: ParsableCommand {
 
   /// Validates if the arguments of this command are valid
   func validate() throws {
-    guard !typeNames.isEmpty else {
-      throw InspectorError.emptyArgument(argumentName: "--type-names")
-    }
```
fdiaz, line 58: Same as https://github.com/fdiaz/SwiftInspector/pull/47/commits/1ab2e779475a3b983f5dcf16de2dbc58b8ba2154#r562985606

Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
```diff
@@ -41,12 +41,12 @@ final class InitializerCommandSpec: QuickSpec {
...
       context("when path is invalid") {
         it("fails when empty") {
-          let result = try? TestTask.run(withArguments: ["initializer", "--path", ""])
+          let result = try? TestTask.run(withArguments: ["initializer", "--path", "", "--name", "FakeName"])
           expect(result?.didFail) == true
         }
 
         it("fails when it doesn't exist") {
-          let result = try? TestTask.run(withArguments: ["initializer", "--path", "/fake/path"])
+          let result = try? TestTask.run(withArguments: ["initializer", "--path", "/fake/path", "--name", "FakeName"])
```
fdiaz, line 49: I noticed these 2 tests were not actually testing what they said because name was empty, they'd fail on that validation instead of here.

Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
```diff
@@ -87,7 +87,7 @@ final class PropertyAnalyzerCommandSpec: QuickSpec {
         }
 
         it("fails when path is empty") {
-          let result = try? TestTask.run(withArguments: ["properties", "--name", "Some"])
+          let result = try? TestTask.run(withArguments: ["properties", "--name", "Some", "--path", ""])
```
fdiaz, line 90: This test was not testing what it said, we were never passing an empty path here.

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     # Checks-out the repo. More at: https://github.com/actions/checkout
     - uses: actions/checkout@v2
     - name: Select Xcode version
-      run: sudo xcode-select -switch /Applications/Xcode_11.4.app
+      run: sudo xcode-select -switch /Applications/Xcode_12.app
```
dfed, line 12: are we considering this a breaking change in our release versioning?

Package.resolved
```diff
@@ -1,58 +1,40 @@
...
       },
       {
         "package": "swift-argument-parser",
         "repositoryURL": "https://github.com/apple/swift-argument-parser",
         "state": {
           "branch": null,
-          "revision": "8d31a0905c346a45c87773ad50862b5b3df8dff6",
-          "version": "0.0.4"
+          "revision": "92646c0cdbaca076c8d3d0207891785b3379cbff",
+          "version": "0.3.1"
```
dfed, line 28: 🙌

Sources/SwiftInspectorCommands/Tests/InitializerCommandSpec.swift
dfed: oops. could we validate that the exception thrown is the one we want?

Sources/SwiftInspectorCommands/Tests/PropertyAnalyzerCommandSpec.swift
dfed: same q here: could we assert that the thrown error is what we expect?

dfed (approved):

.github/workflows/ci.yml
fdiaz: @dfed we are but we're still pre-production https://github.com/fdiaz/SwiftInspector/tags

fdiaz:
> could we validate that the exception thrown is the one we want

We can't. These tests are _integration_ tests, i.e. they talk to the binary directly, so we're not in Swift territory anymore. See https://github.com/fdiaz/SwiftInspector/blob/aba9c842c01905cdb672aff3153fcbec7807a412/Sources/SwiftInspectorCommands/Tests/TestTask.swift#L27 for more on how these works under the hood.

@dfed

fdiaz:
Well maybe we could actually by checking the error message instead?

I'm gonna merge this one and see if I can make that work, interesting idea!

.github/workflows/ci.yml
dfed: 😅 yes I know but this should likely be a feature bump given that we're pre-prod

PR #52, opened 2021-01-27
Create test library and introduce more test helpers

### LINKED ISSUE #14: Add a testing utility library
Right now the testing utilities `TestTask` and `Temporary.swift` are part of different unrelated libraries. We should split them into its own testing library so we import that library only in tests.

dfed:
Resolves #14. Best reviewed commit-by-commit. I haven't yet moved `TestTask` to this module since it doesn't have a spec file.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -37,6 +37,7 @@ let package = Package(
       name: "SwiftInspectorCommandsTests",
       dependencies: [
         "SwiftInspectorCommands",
+        "SwiftInspectorTestHelpers",
         "Nimble",
         "Quick",
     ], path: "Sources/SwiftInspectorCommands/Tests"),
@@ -52,10 +53,23 @@ let package = Package(
       name: "SwiftInspectorAnalyzersTests",
       dependencies: [
         "SwiftInspectorAnalyzers",
+        "SwiftInspectorTestHelpers",
         "SwiftInspectorVisitors",
         "Nimble",
         "Quick",
     ], path: "Sources/SwiftInspectorAnalyzers/Tests"),
+
+    .target(
+      name: "SwiftInspectorTestHelpers",
+      dependencies: ["SwiftSyntax"],
+      exclude: ["Tests"]),
+    .testTarget(
+      name: "SwiftInspectorTestHelpersTests",
+      dependencies: [
+        "SwiftInspectorTestHelpers",
+        "Nimble",
+        "Quick",
+    ], path: "Sources/SwiftInspectorTestHelpers/Tests"),
 
     .target(
       name: "SwiftInspectorVisitors",
@@ -64,6 +78,7 @@ let package = Package(
     .testTarget(
       name: "SwiftInspectorVisitorsTests",
       dependencies: [
+        "SwiftInspectorTestHelpers",
         "SwiftInspectorVisitors",
         "Nimble",
         "Quick",
--- a/Sources/SwiftInspectorAnalyzers/CachedSyntaxTree.swift
+++ b/Sources/SwiftInspectorAnalyzers/CachedSyntaxTree.swift
@@ -47,7 +47,7 @@ public final class CachedSyntaxTree {
   /// 
   /// - Parameter fileURL: The location of the Swift file to parse
   private func memoizeSyntaxTree(at fileURL: URL) throws -> SourceFileSyntax {
-    let cached =  try SyntaxParser.parse(fileURL)
+    let cached = try SyntaxParser.parse(fileURL)
     cachedSyntax[fileURL] = cached
     return cached
   }
--- a/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
+++ b/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
@@ -30,88 +30,67 @@ import SwiftSyntax
 @testable import SwiftInspectorTestHelpers
 
 final class VisitorExecutorSpec: QuickSpec {
+
   private final class MockVisitor: SyntaxVisitor {
-    var ifSyntaxVisitCount = 0
+    var ifStatementSyntaxVisitCount = 0
     override func visitPost(_ node: IfStmtSyntax) {
-      ifSyntaxVisitCount += 1
+      ifStatementSyntaxVisitCount += 1
     }
   }
+
   private final class MockRewriter: SyntaxRewriter {
-    var ifSyntaxVisitCount = 0
+    var ifStatementSyntaxVisitCount = 0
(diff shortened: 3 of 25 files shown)


Package.swift
```diff
@@ -52,18 +53,32 @@ let package = Package(
       name: "SwiftInspectorAnalyzersTests",
       dependencies: [
         "SwiftInspectorAnalyzers",
+        "SwiftInspectorTestHelpers",
         "SwiftInspectorVisitors",
         "Nimble",
         "Quick",
     ], path: "Sources/SwiftInspectorAnalyzers/Tests"),
 
+    .target(
```
dfed, line 62: Note that this target doesn't have an associated library declared at the top of this package. I believe that means we aren't exposing this library externally.

Sources/SwiftInspectorAnalyzers/CachedSyntaxTree.swift
```diff
@@ -47,7 +47,7 @@ public final class CachedSyntaxTree {
   /// 
   /// - Parameter fileURL: The location of the Swift file to parse
   private func memoizeSyntaxTree(at fileURL: URL) throws -> SourceFileSyntax {
-    let cached =  try SyntaxParser.parse(fileURL)
```
dfed, line 50: giphy

Sources/SwiftInspectorTestHelpers/VisitorExecutor.swift
```diff
@@ -0,0 +1,66 @@
...
+  /// Creates a Swift file in the temporary directory with the provided content, and then walks the provided syntax rewriter along the content's syntax.
+  ///
+  /// - Parameters:
+  ///   - content: The content to turn into source and walk.
+  ///   - rewriter: The syntax rewriter to walk along the content syntax.
+  /// - Returns: The location of the file with content written to it.
+  /// - Note: Use a visitor when possible. Rewriters should be used to work around this bug: https://bugs.swift.org/browse/SR-11591
+  public static func createFile<Rewriter: SyntaxRewriter>(
+    withContent content: String,
+    andWalk rewriter: Rewriter)
```
dfed, line 57: Technically this should be `andVisit` but I'm currently optimizing for being able to change rewriters to visitors while changing as few lines as possible. Not sure that's the right thing to optimize for, but that's the context for this decision.

Package.swift
fdiaz: Yup, that's correct

> Library
>
> A library product is used to vend library targets containing the public APIs that will be available to clients.

https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html

Sources/SwiftInspectorAnalyzers/CachedSyntaxTree.swift
fdiaz: I really need to add linter to this project don't I 😁

Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
```diff
@@ -0,0 +1,117 @@
...
+import Foundation
+import Nimble
+import Quick
+import SwiftSyntax
+
+@testable import SwiftInspectorTestHelpers
+
+final class VisitorExecutorSpec: QuickSpec {
+  private final class MockVisitor: SyntaxVisitor {
+    var ifSyntaxVisitCount = 0
```
fdiaz, line 34: `if` and not `on`?

Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
```diff
@@ -0,0 +1,117 @@
...
+    override func visit(_ node: IfStmtSyntax) -> StmtSyntax {
+      ifSyntaxVisitCount += 1
+      return super.visit(node)
+    }
+  }
+
+  override func spec() {
+    describe("createFile(withContent:andWalk:) with visitor") {
+      it("creates a file") {
```
fdiaz, line 49: Starting on Quick 3.0 (which we are on this project) `its` are throwing https://github.com/Quick/Quick/pull/680

So you can simply call `try` directly instead of manually catching the error and failing.

Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
```diff
@@ -0,0 +1,117 @@
...
+  private final class MockRewriter: SyntaxRewriter {
+    var ifSyntaxVisitCount = 0
+    override func visit(_ node: IfStmtSyntax) -> StmtSyntax {
+      ifSyntaxVisitCount += 1
+      return super.visit(node)
+    }
+  }
+
+  override func spec() {
+    describe("createFile(withContent:andWalk:) with visitor") {
```
fdiaz, line 48: `with visitor` should be below and in a `context` block:

```swift

describe("createFile(withContent:andWalk:)") {
  context("with a visitor") { ... }
  context("with a rewriter") { ... }

File: Sources/SwiftInspectorTestHelpers/VisitorExecutor.swift (hunk 3 above)
Comment (fdiaz) [reply Sources/SwiftInspectorTestHelpers/VisitorExecutor.swift:57]: Yeah, there's an open bug on Visitors that don't work properly with the `visit`, so I used Rewriters before knowing. I could use `walk` instead. I'm hoping we can remove all Rewriter to Visitors at one point so that's the right direction 👍

File: codecov.yml [hunk 7]
```diff
@@ -5,6 +5,7 @@ coverage:
       default:
         threshold: 0.1%
 ignore:
-  - "Sources/SwiftInspectorCommands/Tests"
   - "Sources/SwiftInspectorAnalyzers/Tests"
+  - "Sources/SwiftInspectorCommands/Tests"
```
fdiaz, line 9: Thanks for sorting! 👍

fdiaz (approved):

Sources/SwiftInspectorAnalyzers/CachedSyntaxTree.swift
dfed: I wonder which linter we'd use 😉

Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
dfed: it the count of visits to `if` syntax nodes. Will rename to `ifStatementSyntaxVisitCount` to make it a bit more clear

fdiaz: well... that makes a ton of sense

PR #53, opened 2021-01-27
Create GenericRequirementVisitor and spec

dfed:
Creates a visitor capable of finding and reporting generic requirements

### DIFF
--- a/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
+++ b/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
@@ -47,46 +47,20 @@ final class VisitorExecutorSpec: QuickSpec {
   }
 
   override func spec() {
-    describe("createFile(withContent:andWalk:)") {
+    describe("walkVisitor(_:overContent:)") {
       context("with a visitor") {
-        it("creates a file") {
-          let savedURL = try VisitorExecutor.createFile(withContent: "abc", andWalk: MockVisitor())
-          expect(FileManager.default.fileExists(atPath: savedURL.path)) == true
-        }
-
-        it("saves the correct content") {
-          let content = "protocol Some { }"
-          let savedURL = try VisitorExecutor.createFile(withContent: content, andWalk: MockVisitor())
-
-          let savedContent = try String(contentsOf: savedURL, encoding: .utf8)
-          expect(savedContent) == "protocol Some { }"
-        }
-
         it("walks the visitor over the content") {
           let visitor = MockVisitor()
-          _ = try VisitorExecutor.createFile(withContent: "if true {}", andWalk: visitor)
+          try VisitorExecutor.walkVisitor(visitor, overContent: "if true {}")
 
           expect(visitor.ifStatementSyntaxVisitCount) == 1
         }
       }
 
       context("with a rewriter") {
-        it("creates a file") {
-          let savedURL = try VisitorExecutor.createFile(withContent: "abc", andWalk: MockRewriter())
-          expect(FileManager.default.fileExists(atPath: savedURL.path)) == true
-        }
-
-        it("saves the correct content") {
-          let content = "protocol Some { }"
-          let savedURL = try VisitorExecutor.createFile(withContent: content, andWalk: MockRewriter())
-
-          let savedContent = try String(contentsOf: savedURL, encoding: .utf8)
-          expect(savedContent) == "protocol Some { }"
-        }
-
         it("walks the rewriter over the content") {
           let visitor = MockRewriter()
-          _ = try VisitorExecutor.createFile(withContent: "if true {}", andWalk: visitor)
+          try VisitorExecutor.walkVisitor(visitor, overContent: "if true {}")
 
           expect(visitor.ifStatementSyntaxVisitCount) == 1
         }
--- a/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
+++ b/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
@@ -47,7 +47,7 @@ final class VisitorExecutorSpec: QuickSpec {
   }
 
   override func spec() {
-    describe("walkVisitor(_:overContent:)") {
+    describe("walk(:)") {
       context("with a visitor") {
         it("walks the visitor over the content") {
           let visitor = MockVisitor()
--- a/Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
@@ -1,0 +1,82 @@
+// Created by Dan Federman on 1/26/21.
+//
+// Copyright © 2021 Dan Federman
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import SwiftSyntax
+
+public final class GenericRequirementVisitor: SyntaxVisitor {
+
+  public var genericRequirements = [GenericRequirement]()
+
+  public override func visit(_ node: SameTypeRequirementSyntax) -> SyntaxVisitorContinueKind {
+    genericRequirements.append(GenericRequirement(from: node))
+    return .visitChildren
+  }
+
+  public override func visit(_ node: ConformanceRequirementSyntax) -> SyntaxVisitorContinueKind {
+    genericRequirements.append(GenericRequirement(from: node))
+    return .visitChildren
+  }
+
+}
+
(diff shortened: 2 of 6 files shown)


Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public override func visit(_ node: ConformanceRequirementSyntax) -> SyntaxVisitorContinueKind {
+    genericRequirements.append(GenericRequirement(from: node))
+    return .visitChildren
+  }
+
+}
+
+public struct GenericRequirement: Codable, Equatable {
```
dfed, line 44: This type doesn't _need_ to be Codable, but I figure it might be a good thing to make these types capable of being persisted easily.

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -0,0 +1,82 @@
...
+    leftType: String,
+    rightType: String,
+    relationship: Relationship)
+  {
+    self.leftType = leftType
+    self.rightType = rightType
+    self.relationship = relationship
+  }
+
+  init(from node: SameTypeRequirementSyntax) {
```
fdiaz, line 58:
```suggestion
  init(node: SameTypeRequirementSyntax) {
```

> The first argument to initializer and factory methods calls should not form a phrase starting with the base name, e.g. x.makeWidget(cogCount: 47)

https://swift.org/documentation/api-design-guidelines#init-factory-phrase-ends-with-basename

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -0,0 +1,82 @@
...
+    leftType = node.leftTypeIdentifier.description.trimmingCharacters(in: .whitespacesAndNewlines)
+    rightType = node.rightTypeIdentifier.description.trimmingCharacters(in: .whitespacesAndNewlines)
+    relationship = .equals
+  }
+
+  init(from node: ConformanceRequirementSyntax) {
```
fdiaz, line 64:
```suggestion
  init(node: ConformanceRequirementSyntax) {
```

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -0,0 +1,82 @@
...
+    relationship: Relationship)
+  {
+    self.leftType = leftType
+    self.rightType = rightType
+    self.relationship = relationship
+  }
+
+  init(from node: SameTypeRequirementSyntax) {
+    leftType = node.leftTypeIdentifier.description.trimmingCharacters(in: .whitespacesAndNewlines)
+    rightType = node.rightTypeIdentifier.description.trimmingCharacters(in: .whitespacesAndNewlines)
```
fdiaz, line 60: I don't think we should use `description` here. IIUC `description` is just a representation of the instance (from CustomStringConvertible). The fact we're relying on that is a hack IMO that can break unexpectedly since there's no contract that states that this needs to be the same value as the source (e.g. we can imagine that this can change to be more user friendly by providing more information). 

I'd recommend iterating on the possible type of the identifier instead as how we do here https://github.com/fdiaz/SwiftInspector/blob/ae1ef42775efbdb85b9cbd5eb25aa49a5d169c34/Sources/SwiftInspectorAnalyzers/TypeConformanceAnalyzer.swift#L66

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -0,0 +1,252 @@
...
+      self.sut = GenericRequirementVisitor()
+    }
+    afterEach {
+      guard let fileURL = self.fileURL else {
+        return
+      }
+      try? Temporary.removeItem(at: fileURL)
+    }
+
+    describe("visiting a syntax tree involving a protocol") {
```
fdiaz, line 47: Let's follow the previous convention of using the name of the method we're testing here and then `context` for the environment we're testing in 🙏

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -0,0 +1,252 @@
...
+      context("that has no generic constraints") {
+        it("finds no generic requirements") {
+          let content = """
+            public protocol SomeGenericProtocol: GenericProtocol {}
+            """
+
+          self.fileURL = try VisitorExecutor.createFile(
+            withContent: content,
+            andWalk: self.sut)
```
fdiaz, line 56: I'm actually not in love with this method now that I see it in use. This makes it a lot harder to understand what we're actually testing. How do you feel about this method returning both the fileURL and the SourceFileSyntax so we can call `walk` on the visitor here instead, which would make it obvious what we're actually testing and how it should be used?

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -0,0 +1,252 @@
...
+
+          expect(self.sut.genericRequirements.first).to(beNil())
```
fdiaz, line 58:
```suggestion
          expect(self.sut.genericRequirements).to(beEmpty())
```

seems a bit more clear IMHO

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -0,0 +1,82 @@
...
+  // MARK: Public
+
+  public let leftType: String
+  public let rightType: String
+  public let relationship: Relationship
+
+  // MARK: - Relationship
+
+  public enum Relationship: Int, Codable {
+    case equals = 1
```
fdiaz, line 79: Curious, why start at 1 and. not default to the implicit 0?

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -0,0 +1,252 @@
...
+          expect(self.sut.genericRequirements.first) == GenericRequirement(
+            leftType: "LeftType",
+            rightType: "RightType",
+            relationship: .equals)
+        }
+      }
+
+      context("that has two generic constraints") {
+        it("finds both requirements") {
+          let content = """
```
fdiaz, line 83: You can move this "up" to the context and then we can have 2 tests here instead of 1, one for the first type and one for the last type.

fdiaz (commented):
Really cool stuff! One major comment re: the usage of `description` and some nits about tests

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
dfed: because in Objective-C `nil` and `0` are hardish to distinguish, so I always make int-backed enums start with 1 to make it really really clear when there's no data.

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
dfed: ah so we'd use a beforeEach? That'll work!

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -0,0 +1,82 @@
...
+
+import Foundation
+import SwiftSyntax
+
+public final class GenericRequirementVisitor: SyntaxVisitor {
+
+  public var genericRequirements = [GenericRequirement]()
+
+  public override func visit(_ node: SameTypeRequirementSyntax) -> SyntaxVisitorContinueKind {
+    genericRequirements.append(GenericRequirement(from: node))
```
dfed, line 33:
```suggestion
    genericRequirements.append(GenericRequirement(node: node))
```

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -0,0 +1,82 @@
...
+    return .visitChildren
+  }
+
+  public override func visit(_ node: ConformanceRequirementSyntax) -> SyntaxVisitorContinueKind {
+    genericRequirements.append(GenericRequirement(from: node))
```
dfed, line 38:
```suggestion
    genericRequirements.append(GenericRequirement(node: node))
```

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
dfed: Going to try this suggestion out. That said, the [doc comment on `description`](https://github.com/apple/swift-syntax/blob/577d3bbe3a4f39ba8843093f3bd1ae0b593da557/Sources/SwiftSyntax/Syntax.swift#L435) does make it seem like this is a reasonable API to use:
> A source-accurate description of this node.

dfed: 6be54ac145af0dcc587e65717e40f1c9248bc438

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
dfed: f51b1a1577db7d1db7c29457df86fba9ff065531

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
dfed: 41c1ddafbcc3a3bf630401ee1d08f64344f520f6

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
dfed: 41c1ddafbcc3a3bf630401ee1d08f64344f520f6

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
dfed: and a bit of a painful followup 8db6d24fd31072097188d7d5c562afb9d4ba0c36

dfed:
@fdiaz this PR is ready for another review

Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
```diff
@@ -47,46 +47,20 @@ final class VisitorExecutorSpec: QuickSpec {
   }
 
   override func spec() {
-    describe("createFile(withContent:andWalk:)") {
+    describe("walkVisitor(_:overContent:)") {
```
fdiaz, line 50:
```suggestion
    describe("walk(:)") {
```

since we're actually not testing the `walkVisitor` method, but the underlying walk method.

fdiaz (approved):
Thanks!

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
dfed: I was wrong! See #68

PR #54, opened 2021-01-27
ImportsAnalyzer -> ImportSyntaxReader. Also create StandardAnalyzer

dfed:
Migrates the import analyzer underlying tooling down into the Visitor module. Deletes the remainder of `ImportAnalyzer` in favor of a new `StandardAnalyzer`

### DIFF
--- a/Sources/SwiftInspectorAnalyzers/ImportsAnalyzer.swift
+++ b/Sources/SwiftInspectorAnalyzers/ImportsAnalyzer.swift
@@ -1,0 +1,35 @@
+// Created by Dan Federman on 1/28/21.
+//
+// Copyright © 2021 Dan Federman
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import SwiftInspectorVisitors
+
+extension StandardAnalyzer {
+
+  public func analyzeImports(fileURL: URL) throws -> [ImportStatement] {
+    let visitor = ImportSyntaxReader()
+    try analyze(fileURL: fileURL, withVisitor: visitor)
+    return visitor.imports
+  }
+}
--- a/Sources/SwiftInspectorAnalyzers/StandardAnalyzer.swift
+++ b/Sources/SwiftInspectorAnalyzers/StandardAnalyzer.swift
@@ -37,7 +37,7 @@ public struct StandardAnalyzer {
   /// - Parameters:
   ///   - fileURL: The fileURL where the Swift file is located
   ///   - visitor: A Swift syntax visitor to use to analyze the provided file
-  public func analyze<Visitor: SyntaxVisitor>(fileURL: URL, withVisitor visitor: Visitor) throws
+  func analyze(fileURL: URL, withVisitor visitor: SyntaxVisitor) throws
   {
     let syntax: SourceFileSyntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
     visitor.walk(syntax)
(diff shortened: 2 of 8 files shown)


dfed:
I don't get the CodeCov complaint... the method it says we aren't hitting, we are hitting (and I have the breakpoints to prove it!)

Sources/SwiftInspectorAnalyzers/StandardAnalyzer.swift
```diff
@@ -0,0 +1,61 @@
...
+  /// - Parameter cachedSyntaxTree: The cached syntax tree to return the AST tree from
+  public init(cachedSyntaxTree: CachedSyntaxTree = .init()) {
+    self.cachedSyntaxTree = cachedSyntaxTree
+  }
+
+  /// Analyzes a Swift file with the provided visitor
+  /// - Parameters:
+  ///   - fileURL: The fileURL where the Swift file is located
+  ///   - visitor: A Swift syntax visitor to use to analyze the provided file
+  public func analyze<Visitor: SyntaxVisitor>(fileURL: URL, withVisitor visitor: Visitor) throws
```
fdiaz, line 40: I think this class should be internal since we don't wanna expose `SyntaxVisitor` to the Command?

Sources/SwiftInspectorAnalyzers/StandardAnalyzer.swift
```diff
@@ -0,0 +1,61 @@
...
+    let syntax: SourceFileSyntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
+    visitor.walk(syntax)
+  }
+
+  /// Analyzes a Swift file with the provide visitor
+  /// - Parameters:
+  ///   - fileURL: The fileURL where the Swift file is located
+  ///   - visitor: A Swift syntax rewriter to use to analyze the provided file
+  /// - Note: Use a visitor when possible. Rewriters should be used to work around this bug: https://bugs.swift.org/browse/SR-11591
+  public func analyze<Visitor: SyntaxRewriter>(fileURL: URL, withVisitor visitor: Visitor) throws
```
fdiaz, line 51: How are we passing the output from this to the Command to be put to stdout?

Sources/SwiftInspectorCommands/ImportsCommand.swift
```diff
@@ -41,13 +42,14 @@ final class ImportsCommand: ParsableCommand {
...
-    let analyzer = ImportsAnalyzer(cachedSyntaxTree: cachedSyntaxTree)
+    let analyzer = StandardAnalyzer(cachedSyntaxTree: cachedSyntaxTree)
     let fileURL = URL(fileURLWithPath: path)
 
     let outputArray = try FileManager.default.swiftFiles(at: fileURL)
       .reduce(Set<String>()) { result, url in
-        let importStatements = try analyzer.analyze(fileURL: url)
-        let output = importStatements.map { outputString(from: $0) }
+        let reader = ImportSyntaxReader()
+        try analyzer.analyze(fileURL: url, withVisitor: reader)
```
fdiaz, line 51: I'm not sure about this change. This is making `ImportSyntaxReader` which is a subclass of `SyntaxRewriter` visible to the Command, making the boundary a lot more blurry

How do you feel about having a factory that creates these analyzers internally in the Analyzer layer and then we keep the `analyze` method just taking a fileURL?

That way we can make the StandardAnalyzer usage of SwiftSyntax information internal.

Sources/SwiftInspectorCommands/ImportsCommand.swift
```diff
@@ -41,13 +42,14 @@ final class ImportsCommand: ParsableCommand {
...
+        let output = reader.imports.map { outputString(from: $0) }
```
fdiaz, line 52: I don't really love this pattern either, I think it's weird that we need to pass the reader to another class and then _read from the reader_ rather than the Analyzer outputting this information directly.

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -27,7 +27,7 @@ import SwiftSyntax
 
 public final class GenericRequirementVisitor: SyntaxVisitor {
 
-  public var genericRequirements = [GenericRequirement]()
+  public private(set) var genericRequirements = [GenericRequirement]()
```
fdiaz, line 30: 👍

fdiaz (commented):
One comment, everything else looks great though

Sources/SwiftInspectorAnalyzers/StandardAnalyzer.swift
dfed: f6fad8b51e4f0ff3fea59cbdc13f728e1eab1231

Sources/SwiftInspectorCommands/ImportsCommand.swift
dfed: f6fad8b51e4f0ff3fea59cbdc13f728e1eab1231

Sources/SwiftInspectorAnalyzers/ImportsAnalyzer.swift
```diff
@@ -23,130 +23,13 @@
...
-import SwiftSyntax
+import SwiftInspectorVisitors
 
-public final class ImportsAnalyzer: Analyzer {
+extension StandardAnalyzer {
 
-  /// - Parameter cachedSyntaxTree: The cached syntax tree to return the AST tree from
-  public init(cachedSyntaxTree: CachedSyntaxTree = .init()) {
-    self.cachedSyntaxTree = cachedSyntaxTree
+  public func analyzeImports(fileURL: URL) throws -> [ImportStatement] {
```
fdiaz, line 30: I was thinking more of using something like:

```swift
public struct StandardAnalyzer {
  init(cachedSyntaxTree: CachedSyntaxTree, visitor: SyntaxVisitor) { ... }
```

where then you'd have this simply be:

```swift
extension StandardAnalyzer {
  static func imports(cachedSyntaxTree: CachedSyntaxTree = .init()) -> StandardAnalyzer {
    StandardAnalyzer(cachedSyntaxTree: cachedSyntaxTree, visitor: ImportSyntaxReader())
  }
}
```

and keep the `public func analyze(fileURL: URL) throws -> Output` signature on the Analyzer.

This will require us a bit more refactoring since we'd need for each visitor to have a generic returning type (so we can know what `Output` is, so I'm fine doing this in the meantime and we can in the future migrate.

fdiaz (approved):

Sources/SwiftInspectorAnalyzers/ImportsAnalyzer.swift
dfed: Oooooooh. I like this idea. Let's circle back to this though since I'm not convinced we want a single output type on each visitor. Maybe we do... but I want to spend a little more time with the visitors before I commit to that.

PR #55, opened 2021-01-28
Create TypeConformanceVisitor

dfed:
Creates a `TypeConformanceVisitor` an associated spec. This visitor collects conformance information with no additional logic. Since the `TypeConformanceAnalyzer` has significantly more involved logic I haven't updated it. This type will be a building block for a future `TypeDeclarationVisitor`.

The future `TypeDeclarationVisitor` will utilize both `TypeConformanceVisitor` and `GenericRequirementVisitor`, and will return types that it found, as well as the conformance and generic requirements set up in those types. But that's a future PR. You could also imagine this type being used by separate `StructDeclarationVisitor`, `ClassDeclarationVisitor`, etc. But again, I'm getting ahead of myself.

### DIFF
--- a/Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
@@ -33,6 +33,9 @@ final class TypeConformanceVisitorSpec: QuickSpec {
   private var sut = TypeConformanceVisitor()
   
   override func spec() {
+    beforeEach {
+      self.sut = TypeConformanceVisitor()
+    }
 
     describe("visit(_:)") {
       context("when a type conforms to a protocol") {
@@ -41,7 +44,6 @@ final class TypeConformanceVisitorSpec: QuickSpec {
             let content = """
             class SomeObject: Foo {}
             """
-            self.sut = TypeConformanceVisitor()
             try? VisitorExecutor.walkVisitor(self.sut, overContent: content)
           }
 
@@ -55,7 +57,6 @@ final class TypeConformanceVisitorSpec: QuickSpec {
             let content = """
             class SomeObject: Foo, Bar, FooBar {}
             """
-            self.sut = TypeConformanceVisitor()
             try? VisitorExecutor.walkVisitor(self.sut, overContent: content)
           }
 
@@ -70,7 +71,6 @@ final class TypeConformanceVisitorSpec: QuickSpec {
             class SomeObject: Foo, Bar,
               FooBar {}
             """
-            self.sut = TypeConformanceVisitor()
             try? VisitorExecutor.walkVisitor(self.sut, overContent: content)
           }
 
@@ -87,7 +87,6 @@ final class TypeConformanceVisitorSpec: QuickSpec {
             struct SomeStruct: Foo {}
           }
           """
-          self.sut = TypeConformanceVisitor()
           try? VisitorExecutor.walkVisitor(self.sut, overContent: content)
         }
 
@@ -103,7 +102,6 @@ final class TypeConformanceVisitorSpec: QuickSpec {
             struct SomeStruct: Foo {}
           }
           """
-          self.sut = TypeConformanceVisitor()
           try? VisitorExecutor.walkVisitor(self.sut, overContent: content)
         }
 
--- a/Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
@@ -1,0 +1,116 @@
+// Created by Dan Federman on 1/27/21.
+//
+// Copyright © 2021 Dan Federman
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
```diff
@@ -0,0 +1,116 @@
...
+
+import Nimble
+import Quick
+import Foundation
+import SwiftInspectorTestHelpers
+
+@testable import SwiftInspectorVisitors
+
+final class TypeConformanceVisitorSpec: QuickSpec {
+  private var sut = TypeConformanceVisitor()
```
fdiaz, line 33: Let's add this in the `beforeEach` please 🙏 so we get a fresh instance each time

Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
```diff
@@ -0,0 +1,116 @@
...
+  override func spec() {
+
+    describe("visit(_:)") {
+      context("when a type conforms to a protocol") {
+        context("with only one conformance") {
+          beforeEach {
+            let content = """
+            class SomeObject: Foo {}
+            """
+            self.sut = TypeConformanceVisitor()
```
fdiaz, line 44: Let's move this up a level to a beforeEach under the `describe` method

Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
```diff
@@ -0,0 +1,116 @@
...
+          self.sut = TypeConformanceVisitor()
+          try? VisitorExecutor.walkVisitor(self.sut, overContent: content)
+        }
+
+        it("does not find the inner type's conformance the conforming type names") {
+          expect(self.sut.conformsToTypes).to(beEmpty())
+        }
+      }
+
+      context("when both an outer and inner type conform to a protocol") {
```
fdiaz, line 99: Isn't this incorrect for this use case?

Sources/SwiftInspectorVisitors/TypeConformanceVisitor.swift
```diff
@@ -0,0 +1,41 @@
...
+
+  public private(set) var conformsToTypes = [String]()
+
+  public override func visit(_ node: InheritedTypeSyntax) -> SyntaxVisitorContinueKind {
+    conformsToTypes.append(node.typeName.description.trimmingCharacters(in: .whitespacesAndNewlines))
+    return .visitChildren
+  }
+
+  public override func visit(_ node: MemberDeclBlockSyntax) -> SyntaxVisitorContinueKind {
+    // We don't care about the internals of this type.
```
fdiaz, line 38: Is that true for this use case? Shouldn't we care here?

Sources/SwiftInspectorVisitors/TypeConformanceVisitor.swift
```diff
@@ -0,0 +1,41 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import SwiftSyntax
+
+public final class TypeConformanceVisitor: SyntaxVisitor {
```
fdiaz, line 28: Let's name this `TypeInheritance`. I think that's more correct than the type conformance that I used 😁

fdiaz (approved):
Approving since this is still a move in the right direction but I think we'd want to find all the type inheritance clauses inside too here?

Sources/SwiftInspectorVisitors/TypeConformanceVisitor.swift
dfed: What clause would this visitor care about within this code block?

As an example, say we're using this type inside of a `ClassDeclSyntax` block to find the information about a class definition. We don't care about what's in the class itself, we just care about the declaration. If there's another class defined within this class, it's up to the parent visitor to continue walking the tree and then invoke this visitor again if we run into another class.

dfed:
> I think we'd want to find all the type inheritance clauses inside too here?

@fdiaz can you give me an example of the information we're not currently grabbing? I'm pretty sure we are grabbing type inheritance clauses.

Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
dfed: Do you mean the behavior is incorrect? I don't think so. Similar to what I wrote in https://github.com/fdiaz/SwiftInspector/pull/55#discussion_r565742802, I think we want these visitors to be as narrowly scoped as possible

Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
dfed: dd3547e5838886fcb61168abf5c7a65d03437b9f

Sources/SwiftInspectorVisitors/Tests/TypeConformanceVisitorSpec.swift
dfed: dd3547e5838886fcb61168abf5c7a65d03437b9f

Sources/SwiftInspectorVisitors/TypeConformanceVisitor.swift
dfed: 4c315985c8c12485a73704e3e97a6b6f11b17d49

dfed:
I'll merge when CI is green because I want to keep building on top of this (and future PRs may help explain what I'm trying to describe here). But I do want to continue discussing the unresolved comments.

Sources/SwiftInspectorVisitors/TypeConformanceVisitor.swift
fdiaz: Gotcha! That makes total sense! Thanks!

fdiaz:
@dfed all good! Your explanation makes total sense to me 😉

PR #56, opened 2021-01-28
Create StructVisitor

dfed:
Best reviewed commit-by-commit. It may be easier to review unit tests before reviewing the shipped code.

Creates a `StructVisitor` in the visitors module. This visitor is capable of determining type information about a struct, as well as finding all of the nested struct definitions within that single struct. This type will get significantly more powerful as we build out a `ClassVisitor` and `EnumVisitor` (see `TODO`s in both tests and code). Both `ClassVisitor` and `EnumVisitor` will look very similar to this visitor, so I'd love to get feedback before continuing to build more of these types.

My current thinking is that this `StructVisitor` will eventually be utilized by a `FileVisitor`.

### DIFF
--- a/Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
@@ -31,14 +31,20 @@ public final class GenericRequirementVisitor: SyntaxVisitor {
 
   public override func visit(_ node: SameTypeRequirementSyntax) -> SyntaxVisitorContinueKind {
     genericRequirements.append(GenericRequirement(node: node))
-    return .visitChildren
+    // We only care about siblings, not children.
+    return .skipChildren
   }
 
   public override func visit(_ node: ConformanceRequirementSyntax) -> SyntaxVisitorContinueKind {
     genericRequirements.append(GenericRequirement(node: node))
-    return .visitChildren
+    // We only care about siblings, not children.
+    return .skipChildren
   }
 
+  public override func visit(_ node: MemberDeclBlockSyntax) -> SyntaxVisitorContinueKind {
+    // We don't care about the internals of a type with generic requirements.
+    .skipChildren
+  }
 }
 
 public struct GenericRequirement: Codable, Equatable {
--- a/Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
@@ -31,18 +31,19 @@ public final class GenericRequirementVisitor: SyntaxVisitor {
 
   public override func visit(_ node: SameTypeRequirementSyntax) -> SyntaxVisitorContinueKind {
     genericRequirements.append(GenericRequirement(node: node))
-    // We only care about siblings, not children.
+    // Children don't have any more information about generic requirements, so don't visit them.
     return .skipChildren
   }
 
   public override func visit(_ node: ConformanceRequirementSyntax) -> SyntaxVisitorContinueKind {
     genericRequirements.append(GenericRequirement(node: node))
-    // We only care about siblings, not children.
+    // Children don't have any more information about generic requirements, so don't visit them.
     return .skipChildren
   }
 
   public override func visit(_ node: MemberDeclBlockSyntax) -> SyntaxVisitorContinueKind {
-    // We don't care about the internals of a type with generic requirements.
+    // A member declaration block means we've found the body of the type.
+    // There's nothing in this body that would help us determine generic requirements.
     .skipChildren
   }
 }
--- a/Sources/SwiftInspectorVisitors/StructVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/StructVisitor.swift
@@ -31,6 +31,7 @@ public final class StructVisitor: SyntaxVisitor {
     self.parentTypeName = parentTypeName
   }
 
+  /// All of the structs found by this visitor.
   public var structs: [StructInfo] {
     [structInfo].compactMap { $0 } + innerStructs
   }
--- a/Sources/SwiftInspectorVisitors/StructVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/StructVisitor.swift
@@ -1,0 +1,121 @@
+// Created by Dan Federman on 1/27/21.
+//
+// Copyright © 2021 Dan Federman
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import SwiftSyntax
+
+public final class StructVisitor: SyntaxVisitor {
+
+  public init(parentTypeName: String? = nil) {
+    self.parentTypeName = parentTypeName
+  }
+
+  public var structs: [StructInfo] {
+    [structInfo].compactMap { $0 } + innerStructs
+  }
+
+  // TODO: also find and nested classes
+  // TODO: also find and nested enums
+
+  public override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+
+    guard !hasFinishedParsingStruct else {
+      assertionFailure("Encountered more than one top-level struct. This is a usage error: a single StructVisitor instance should start walking only over a node of type `StructDeclSyntax`")
+      return .skipChildren
+    }
+
+    if let structInfo = structInfo {
+      // Base case. We've previously found a struct declaration, so this must be an inner struct.
+      // This struct visitor shouldn't recurse down into the children.
+      // Instead, we'll use a new struct visitor to get the information from this struct.
+      let qualifiedParentTypeName: String
+      if let parentTypeName = parentTypeName {
+        qualifiedParentTypeName = "\(parentTypeName).\(structInfo.name)"
+      } else {
+        qualifiedParentTypeName = structInfo.name
+      }
+
+      let innerStructVisitor = StructVisitor(parentTypeName: qualifiedParentTypeName)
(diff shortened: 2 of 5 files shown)


Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -180,10 +180,8 @@ final class GenericRequirementVisitorSpec: QuickSpec {
       context("visiting a syntax tree involving an associatedtype that has one generic constraint") {
         it("finds the generic requirements") {
           let content = """
-            public protocol Test {
-              associatedtype SomeType: SomeProtocol where
-                Element: AnyObject
-            }
+            associatedtype SomeType: SomeProtocol where
```
dfed, line 183: now that we are scoping this visitor to only care about where clauses and not visit children in member declarations, we need to focus this content a bit to only provide a bare-bones associated type with a where clause.

The changes below follow this theme.

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+import SwiftSyntax
+
+public final class StructVisitor: SyntaxVisitor {
+
+  public init(parentTypeName: String? = nil) {
+    self.parentTypeName = parentTypeName
+  }
+
+  public var structs: [StructInfo] {
+    [structInfo].compactMap { $0 } + innerStructs
```
dfed, line 35: This line may be a dash too clever. I'm trying to create an array that includes both the top-level struct we've found and the inner structs. The top-level struct is optional, however, which led to the `compactMap` call.

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  // TODO: also find and nested classes
+  // TODO: also find and nested enums
```
dfed, line 39: while we have code to visit classes and enums, we don't yet find details about them or expose that we've found them. This TODO captures that need.

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+      } else {
+        qualifiedParentTypeName = structInfo.name
+      }
+
+      let innerStructVisitor = StructVisitor(parentTypeName: qualifiedParentTypeName)
+      innerStructVisitor.walk(node)
+
+      self.innerStructs += innerStructVisitor.structs
+      // We've already gotten information from the children from our inner struct visitor.
+      return .skipChildren
```
dfed, line 64: in case this is confusing to the reader: this `if` branch is a base case because _this instance_ does not recurse. We do create another instance that does work for us, but this instance stops parsing deeper down the tree in this case.

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+      // We need to get its information and then visit children to see if there is more information we need.
+      let name = node.identifier.text
+      let typeInheritanceVisitor = TypeInheritanceVisitor()
+      typeInheritanceVisitor.walk(node)
+
+      structInfo = StructInfo(
+        name: name,
+        inheritsFromTypes: typeInheritanceVisitor.inheritsFromTypes,
+        parentTypeName: parentTypeName)
+      return .visitChildren
```
dfed, line 77: I called this case the recursive one because this instance continues to recurse down to the child nodes.

The base true base case is "we no longer have nodes to parse"... but when we hit that case we're not in this method 😅

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  public override func visitPost(_ node: StructDeclSyntax) {
+    hasFinishedParsingStruct = node.identifier.text == structInfo?.name
+  }
+
+  public override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+    if let _ = structInfo {
```
dfed, line 86: in a future PR we'll use the value of this `if let`. See TODO below.

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+      // TODO: Utilize class visitor to find inner class information, utilizing parent information from structInfo
+    } else {
+      // We've encountered a class declaration before encountering a struct declaration. Something is wrong.
+      assertionFailure("Encountered a top-level class. This is a usage error: a single StructVisitor instance should start walking only over a node of type `StructDeclSyntax`")
+    }
+    return .skipChildren
+  }
+
+  public override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
+    if let _ = structInfo {
```
dfed, line 97: in a future PR we'll use the value of this `if let`. See TODO below.

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+  private var hasFinishedParsingStruct = false
+  private var structInfo: StructInfo?
+  private var innerStructs = [StructInfo]()
+}
+
+public struct StructInfo: Codable, Equatable {
+  public let name: String
+  public let inheritsFromTypes: [String]
+  public let parentTypeName: String?
+  // TODO: also find and expose properties on a struct
```
dfed, line 120: When we eventually start parsing properties, we'll be able to visit variable declarations and add them to our `structInfo` property. We won't accidentally scoop up variable declarations from inner structs because every instance of `StructVisitor` stops parsing children when it hits another struct.

dfed:
Force pushed the last commit with no changes in order to kick off the build queue. There was a GitHub incident with Github Actions last night.

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -31,14 +31,20 @@ public final class GenericRequirementVisitor: SyntaxVisitor {
 
   public override func visit(_ node: SameTypeRequirementSyntax) -> SyntaxVisitorContinueKind {
     genericRequirements.append(GenericRequirement(node: node))
-    return .visitChildren
+    // We only care about siblings, not children.
```
fdiaz, line 34: Let's say _why_ here, not what. This is quite a cat comment atm 😁

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -31,14 +31,20 @@ public final class GenericRequirementVisitor: SyntaxVisitor {
...
+    return .skipChildren
   }
 
   public override func visit(_ node: ConformanceRequirementSyntax) -> SyntaxVisitorContinueKind {
     genericRequirements.append(GenericRequirement(node: node))
-    return .visitChildren
+    // We only care about siblings, not children.
```
fdiaz, line 40: Let's uncatify this one too.

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
```diff
@@ -31,14 +31,20 @@ public final class GenericRequirementVisitor: SyntaxVisitor {
...
+    return .skipChildren
   }
 
+  public override func visit(_ node: MemberDeclBlockSyntax) -> SyntaxVisitorContinueKind {
+    // We don't care about the internals of a type with generic requirements.
```
fdiaz, line 45: We can probably do a little bit of uncatifying here too?

Sources/SwiftInspectorVisitors/StructVisitor.swift
fdiaz: I think this is fine, but let's add doc comments

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -0,0 +1,121 @@
...
+    [structInfo].compactMap { $0 } + innerStructs
+  }
+
+  // TODO: also find and nested classes
+  // TODO: also find and nested enums
+
+  public override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+
+    guard !hasFinishedParsingStruct else {
+      assertionFailure("Encountered more than one top-level struct. This is a usage error: a single StructVisitor instance should start walking only over a node of type `StructDeclSyntax`")
```
fdiaz, line 44: This is great 👍

Sources/SwiftInspectorVisitors/Tests/StructVisitorSpec.swift
```diff
@@ -0,0 +1,274 @@
...
+                  public struct BarFooFoo2Struct {}
+                }
+              }
+              """
+
+            try VisitorExecutor.walkVisitor(
+              self.sut,
+              overContent: content)
+
+            expect(self.sut.structs) == [
```
fdiaz, line 117: I'd recommend 2 things here

1) divide these into one it per struct
2) don't initialize the struct directly, but instead check their properties:

```swift
expect(someStruct.name) == "FooStruct"
expect(someStruct. inheritsFromTypes) == []
...
```

This way if we were to add 1 property to `StructInfo`, we don't need to modify n tests, we just modify the tests we care about.

Sources/SwiftInspectorVisitors/Tests/StructVisitorSpec.swift
```diff
@@ -0,0 +1,274 @@
...
+        context("with multiple top-level structs") {
+          it("asserts") {
+            let content = """
+            public struct FooStruct {}
+            public struct BarStruct {}
+            """
+
+            // The StructVisitor is only meant to be used over a single struct.
+            // Using a StructVisitor over a block that has multiple top-level
+            // structs is API misuse.
```
fdiaz, line 198: 👍

fdiaz (approved):
Some comments but this looks great to me!

Sources/SwiftInspectorVisitors/TypeInheritanceVisitor.swift
```diff
@@ -31,7 +31,8 @@ public final class TypeInheritanceVisitor: SyntaxVisitor {
 
   public override func visit(_ node: InheritedTypeSyntax) -> SyntaxVisitorContinueKind {
     inheritsFromTypes.append(node.typeName.description.trimmingCharacters(in: .whitespacesAndNewlines))
-    return .visitChildren
+    // We care about sibling types but not children
```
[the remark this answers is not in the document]
dfed: 1caf37d99c1cbd555b1f4e26b08de4eaf18fc800

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
dfed: 1caf37d99c1cbd555b1f4e26b08de4eaf18fc800

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
dfed: 1caf37d99c1cbd555b1f4e26b08de4eaf18fc800

Sources/SwiftInspectorVisitors/GenericRequirementVisitor.swift
dfed: 1caf37d99c1cbd555b1f4e26b08de4eaf18fc800

Sources/SwiftInspectorVisitors/Tests/StructVisitorSpec.swift
dfed: 😬 I don't love the idea of having to access each item in the array individually. But I do like the idea of checking properties... so... okay fine 😉

dfed: e2b5e143a2fc7568a76bf5fa11ba9f6a4db10dd8

Sources/SwiftInspectorVisitors/StructVisitor.swift
dfed: a7a9dc538134133bbfc18e420e441dcf4982f6dc

PR #57, opened 2021-01-28
Migrate SyntaxRewriter to use SyntaxVisitor instead

fdiaz:
When Is started this project I didn't realize I could use the `walk` method to accomplish a similar result than calling `visit` directly. There's an open bug for calling the `visit` method directly.

This PR removes all `SyntaxRewriter` in favor of using `SyntaxVisitor` and the `walk` method instead of `visit`.

This simplifies our reusable code so we don't need to fork so much.

### DIFF
--- a/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
+++ b/Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
@@ -38,32 +38,13 @@ final class VisitorExecutorSpec: QuickSpec {
     }
   }
 
-  private final class MockRewriter: SyntaxRewriter {
-    var ifStatementSyntaxVisitCount = 0
-    override func visit(_ node: IfStmtSyntax) -> StmtSyntax {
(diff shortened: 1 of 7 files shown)


Sources/SwiftInspectorTestHelpers/Tests/VisitorExecutorSpec.swift
```diff
@@ -38,32 +38,13 @@ final class VisitorExecutorSpec: QuickSpec {
...
-        it("walks the rewriter over the content") {
-          let visitor = MockRewriter()
-          try VisitorExecutor.walkVisitor(visitor, overContent: "if true {}")
+      it("walks the visitor over the content") {
+        let visitor = MockVisitor()
+        try VisitorExecutor.walkVisitor(visitor, overContent: "if true {}")
 
-          expect(visitor.ifStatementSyntaxVisitCount) == 1
-        }
+        expect(visitor.ifStatementSyntaxVisitCount) == 1
```
fdiaz, line 47: This just removes the rewriter test.

dfed (approved):
🎉

PR #59, opened 2021-01-28
Create ProtocolVisitor

dfed:
Creates a `ProtocolVisitor` in the visitors module. This visitor is capable of determining type information about a protocol.

### DIFF
--- a/Sources/SwiftInspectorVisitors/ProtocolVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/ProtocolVisitor.swift
@@ -1,0 +1,83 @@
+// Created by Dan Federman on 1/28/21.
+//
+// Copyright © 2021 Dan Federman
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorVisitors/ProtocolVisitor.swift
```diff
@@ -0,0 +1,83 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import SwiftSyntax
+
+public final class ProtocolVisitor: SyntaxVisitor {
+
+  public var protocolInfo: ProtocolInfo?
```
fdiaz, line 30: private set?

fdiaz (approved):
Really cool stuff, I'm really liking how this is building on top of each other now 👍

Sources/SwiftInspectorVisitors/ProtocolVisitor.swift
dfed: yes! how do I keep forgetting this 😂 
```suggestion
  public private(set) var protocolInfo: ProtocolInfo?
```

PR #62, opened 2021-01-29
Create ExtensionVisitor

dfed:
Creates an `ExtensionVisitor` in the visitors module. This visitor is capable of determining type information about an extension.

### DIFF
--- a/Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
@@ -72,9 +72,9 @@ public final class ExtensionVisitor: SyntaxVisitor {
   }
 
   public override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
-    if !hasFinishedParsingExtension, let _ = extensionInfo {
+    if !hasFinishedParsingExtension, let extensionInfo = extensionInfo {
       // We've previously found an extension declaration, so this must be an inner struct.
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
```diff
@@ -0,0 +1,120 @@
...
+      // We've encountered a class declaration before encountering an extension declaration. Something is wrong.
+      assertionFailure("Encountered a top-level class. This is a usage error: a single ExtensionVisitor instance should start walking only over a node of type `ExtensionDeclSyntax`")
+    }
+    return .skipChildren
+  }
+
+  public override func visit(_ node: StructDeclSyntax) -> SyntaxVisitorContinueKind {
+    if !hasFinishedParsingExtension, let _ = extensionInfo {
+      // We've previously found an extension declaration, so this must be an inner struct.
+      let structVisitor = StructVisitor(parentTypeName: extensionInfo?.name)
```
fdiaz, line 77: Why the optionality if we can pass a non-optional?

fdiaz (approved):

Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
dfed: 😂 good catch

dfed: d237eff

PR #65, opened 2021-01-29
Do not rely on ordering in tests

dfed:
Fixing some debt incurred in #56

### DIFF
--- a/Sources/SwiftInspectorVisitors/Tests/StructVisitorSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/StructVisitorSpec.swift
@@ -116,80 +116,73 @@ final class StructVisitorSpec: QuickSpec {
           }
 
           it("finds FooStruct") {
-            guard self.sut.structs.count > 0 else {
-              fail("FooStruct not found at expected index")
-              return
+            let matching = self.sut.structs.filter {
+              $0.name == "FooStruct"
+                && $0.inheritsFromTypes == []
+                && $0.parentTypeName == nil
             }
-            let structInfo = self.sut.structs[0]
-            expect(structInfo.name) == "FooStruct"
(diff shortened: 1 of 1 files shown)


Sources/SwiftInspectorVisitors/Tests/StructVisitorSpec.swift
```diff
@@ -116,80 +116,73 @@ final class StructVisitorSpec: QuickSpec {
...
+              $0.name == "FooStruct"
+                && $0.inheritsFromTypes == []
+                && $0.parentTypeName == nil
             }
-            let structInfo = self.sut.structs[0]
-            expect(structInfo.name) == "FooStruct"
-            expect(structInfo.inheritsFromTypes) == []
-            expect(structInfo.parentTypeName).to(beNil())
+
+            expect(matching.count) == 1
```
dfed, line 125: We can use [this API](https://github.com/Quick/Nimble/blob/0f494fe0021cba742c7369981381603dc8a4c379/Sources/Nimble/Matchers/ContainElementSatisfying.swift) to be more semantic. Since I have a _lot_ of tests in this from in multiple open PRs I'm going to do this change all at once once these PRs merge.

dfed: I'm actually not sure that API does what we're doing here. I think that test is `count >= 1` where I want `count == 1`

dfed: Francisco has suggested in DMs writing a helper method to avoid needing to do the filtering within these tests.

fdiaz: No need to do that right now though 👍

fdiaz (approved):

PR #67, opened 2021-01-30
Enable FileVisitor to find all types

dfed:
This PR builds on top of all the prior PRs to enable the `FileVisitor` to get information about all types defined within a particular file. Note that `ClassVisitor`, `EnumVisitor`, and `ExtensionVisitor` were all built out in parallel PRs, so this PR pulls all those types together.

### DIFF
--- a/Sources/SwiftInspectorAnalyzers/Tests/StaticUsageAnalyzerSpec.swift
+++ b/Sources/SwiftInspectorAnalyzers/Tests/StaticUsageAnalyzerSpec.swift
@@ -184,7 +184,6 @@ final class StaticUsageAnalyzerSpec: QuickSpec {
         let content = """
         JitneyProducer.shared.publish(
               PaidGrowth.V1.PaidGrowthSignupCompletePixelEvent(
-                // TODO: Use the globalLoggingContext property on an injected LoggingService than accessing a singleton directly.
                 context: EventContext.shared.loggingContext(),
                 device_id: trackingManager.value(forAirEventsSharedKey: deviceIDParamKey) ?? "",
                 user_id: Int64(accountManager.activeAccount?.user.userId ?? "") ?? 0
--- a/Sources/SwiftInspectorVisitors/ClassVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/ClassVisitor.swift
@@ -38,8 +38,8 @@ public final class ClassVisitor: SyntaxVisitor {
 
   /// Inner structs found by this visitor.
   public private(set) var innerStructs = [StructInfo]()
-
-  // TODO: also find and nested enums
+  /// Inner enums found by this visitor.
+  public private(set) var innerEnums = [EnumInfo]()
 
   public override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
 
@@ -59,7 +59,7 @@ public final class ClassVisitor: SyntaxVisitor {
       let innerClassVisitor = ClassVisitor(parentTypeName: qualifiedParentTypeName)
       innerClassVisitor.walk(node)
 
-      self.innerClasses += innerClassVisitor.classes
+      innerClasses += innerClassVisitor.classes
       // We've already gotten information from the children from our inner class visitor.
       return .skipChildren
 
@@ -94,6 +94,8 @@ public final class ClassVisitor: SyntaxVisitor {
       structVisitor.walk(node)
       innerStructs += structVisitor.structs
       innerClasses += structVisitor.innerClasses
+      innerEnums += structVisitor.innerEnums
+
     } else {
       // We've encountered a struct declaration before encountering a class declaration. Something is wrong.
       assertionFailure("Encountered a top-level struct. This is a usage error: a single ClassVisitor instance should start walking only over a node of type `ClassDeclSyntax`")
@@ -102,9 +104,18 @@ public final class ClassVisitor: SyntaxVisitor {
   }
 
   public override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
-    if !hasFinishedParsingClass, let _ = classInfo {
+    if !hasFinishedParsingClass, let classInfo = classInfo {
       // We've previously found a class declaration, so this must be an inner enum.
-      // TODO: Utilize enum visitor to find inner enum information, utilizing parent information from structInfo
+      let qualifiedParentTypeName = QualifiedParentNameCreator.createNameGiven(
+        currentParentTypeName: parentTypeName,
+        currentTypeName: classInfo.name)
+
+      let enumVisitor = EnumVisitor(parentTypeName: qualifiedParentTypeName)
+      enumVisitor.walk(node)
+      innerEnums += enumVisitor.enums
+      innerStructs += enumVisitor.innerStructs
+      innerClasses += enumVisitor.innerClasses
+
     } else {
       // We've encountered an enum declaration before encountering a class declaration. Something is wrong.
       assertionFailure("Encountered a top-level enum. This is a usage error: a single ClassVisitor instance should start walking only over a node of type `ClassDeclSyntax`")
--- a/Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
@@ -67,9 +67,9 @@ public final class ExtensionVisitor: SyntaxVisitor {
       let classVisitor = ClassVisitor(parentTypeName: extensionInfo.name)
       classVisitor.walk(node)
 
-      innerClasses.append(contentsOf: classVisitor.classes)
-      innerStructs.append(contentsOf: classVisitor.innerStructs)
-      innerEnums.append(contentsOf: classVisitor.innerEnums)
+      innerClasses += classVisitor.classes
+      innerStructs += classVisitor.innerStructs
+      innerEnums += classVisitor.innerEnums
 
     } else {
       // We've encountered a class declaration before encountering an extension declaration. Something is wrong.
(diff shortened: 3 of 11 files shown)


Sources/SwiftInspectorAnalyzers/Tests/StaticUsageAnalyzerSpec.swift
```diff
@@ -184,7 +184,6 @@ final class StaticUsageAnalyzerSpec: QuickSpec {
         let content = """
         JitneyProducer.shared.publish(
               PaidGrowth.V1.PaidGrowthSignupCompletePixelEvent(
-                // TODO: Use the globalLoggingContext property on an injected LoggingService than accessing a singleton directly.
```
dfed, line 187: This change isn't related to this PR but I was searching for TODOs and this one was distracting 😅

Sources/SwiftInspectorVisitors/ClassVisitor.swift
```diff
@@ -59,7 +59,7 @@ public final class ClassVisitor: SyntaxVisitor {
       let innerClassVisitor = ClassVisitor(parentTypeName: qualifiedParentTypeName)
       innerClassVisitor.walk(node)
 
-      self.innerClasses += innerClassVisitor.classes
```
dfed, line 62: get that lint

bachand:
I will check this out on Monday if that's OK @dfed

dfed:
Zero rush @bachand. Enjoy the weekend 🙂

Sources/SwiftInspectorVisitors/ClassVisitor.swift
```diff
@@ -102,9 +104,18 @@ public final class ClassVisitor: SyntaxVisitor {
...
-      // TODO: Utilize enum visitor to find inner enum information, utilizing parent information from structInfo
+      let qualifiedParentTypeName = QualifiedParentNameCreator.createNameGiven(
+        currentParentTypeName: parentTypeName,
+        currentTypeName: classInfo.name)
+
+      let enumVisitor = EnumVisitor(parentTypeName: qualifiedParentTypeName)
+      enumVisitor.walk(node)
+      innerEnums += enumVisitor.enums
+      innerStructs += enumVisitor.innerStructs
+      innerClasses += enumVisitor.innerClasses
```
bachand, line 117: This matches the code for inner structs.

Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
```diff
@@ -77,7 +84,9 @@ public final class ExtensionVisitor: SyntaxVisitor {
       let structVisitor = StructVisitor(parentTypeName: extensionInfo.name)
       structVisitor.walk(node)
 
-      structs.append(contentsOf: structVisitor.structs)
+      innerStructs.append(contentsOf: structVisitor.structs)
```
bachand, line 87: Nit: in some places we use `+=` and in others we use `append(...)`. Would be nice to be consistent, so that differences have meaning (which I'm not sure these do).

Sources/SwiftInspectorVisitors/FileVisitor.swift
```diff
@@ -72,16 +72,34 @@ public final class FileVisitor: SyntaxVisitor {
...
-    // TODO: append structs
-    // TODO: append classes
+    let enumsVisitor = EnumVisitor()
+    enumsVisitor.walk(node)
+
+    fileInfo.appendEnums(enumsVisitor.enums)
+    fileInfo.appendClasses(enumsVisitor.innerClasses)
+    fileInfo.appendStructs(enumsVisitor.innerStructs)
+
+    // We don't need to visit children because our visitor just did that for us.
```
bachand, line 89: Helpful comment.

Sources/SwiftInspectorVisitors/FileVisitor.swift
```diff
@@ -108,4 +127,10 @@ public struct FileInfo: Codable, Equatable {
   mutating func appendClasses(_ classes: [ClassInfo]) {
     self.classes.append(contentsOf: classes)
   }
+  mutating func appendEnums(_ enums: [EnumInfo]) {
+    self.enums.append(contentsOf: enums)
+  }
+  mutating func appendExtension(_ extensionInfo: ExtensionInfo) {
```
bachand, line 133: Asking to learn: why does this method accept one extension when other methods accept multiple classes and structs?

Sources/SwiftInspectorVisitors/StructVisitor.swift
```diff
@@ -101,9 +106,18 @@ public final class StructVisitor: SyntaxVisitor {
...
-      // TODO: Utilize enum visitor to find inner enum information, utilizing parent information from structInfo
+      let qualifiedParentTypeName = QualifiedParentNameCreator.createNameGiven(
+        currentParentTypeName: parentTypeName,
+        currentTypeName: structInfo.name)
+
+      let enumVisitor = EnumVisitor(parentTypeName: qualifiedParentTypeName)
+      enumVisitor.walk(node)
+      innerEnums += enumVisitor.enums
+      innerClasses += enumVisitor.innerClasses
+      innerStructs += enumVisitor.innerStructs
```
bachand, line 119: It feels like it would be really easy to miss one line here... I wonder if we want a way to generate all valid permutations in our tests to ensure that we cover all cases now and in the future?

bachand: Or maybe it's easy enough to write them all out, as it seems like we are doing now.

Sources/SwiftInspectorVisitors/Tests/ClassVisitorSpec.swift
```diff
@@ -246,6 +246,26 @@ final class ClassVisitorSpec: QuickSpec {
...
+            let matching = self.sut.innerEnums.filter {
+              $0.name == "BarFooEnum"
+                && $0.inheritsFromTypes == []
+                && $0.parentTypeName == "FooClass"
+            }
+
+            expect(matching.count) == 1
+          }
+
+          it("finds BarBarFooClass") {
```
bachand, line 260: This class is the same name as another class nested under a struct. I wonder if we can make each type name in the code we're testing unique, for easier reading of these tests? Right now if you search a type short name you will find tests for different fully qualified types.

bachand: Or we could have the name in the `it(...)` clause be fully qualified but that may get verbose.

Sources/SwiftInspectorVisitors/Tests/FileVisitorSpec.swift
```diff
@@ -131,12 +135,87 @@ final class FileVisitorSpec: QuickSpec {
...
+        it("finds TestClass.InnerEnum") {
+          let matchingStructs = self.sut.fileInfo.enums.filter {
+            $0.name == "InnerEnum"
+              && $0.parentTypeName == "TestClass"
+          }
+          expect(matchingStructs.count) == 1
+        }
+
+        it("finds TestEnum") {
+          let matchingStructs = self.sut.fileInfo.enums.filter {
```
bachand, line 147: Nit: `matchingEnums`. It could be nice to make a name that is applicable to all types of tests to define away this class of copy/paste error.

Sources/SwiftInspectorVisitors/Tests/FileVisitorSpec.swift
```diff
@@ -131,12 +135,87 @@ final class FileVisitorSpec: QuickSpec {
...
           }
           expect(matchingStructs.count) == 1
         }
+
+        it("finds Array extension") {
+          let matchingStructs = self.sut.fileInfo.extensions.filter {
+            $0.name == "Array"
+              && $0.genericRequirements.first?.leftType == "Element"
+              && $0.genericRequirements.first?.rightType == "Int"
+              && $0.genericRequirements.first?.relationship == .equals
```
bachand, line 190: Nice

Sources/SwiftInspectorVisitors/Tests/FileVisitorSpec.swift
```diff
@@ -131,12 +135,87 @@ final class FileVisitorSpec: QuickSpec {
...
+
         it("finds TestProtocol") {
           let matchingStructs = self.sut.fileInfo.protocols.filter {
             $0.name == "TestProtocol"
           }
           expect(matchingStructs.count) == 1
         }
+
+        it("finds Array extension") {
+          let matchingStructs = self.sut.fileInfo.extensions.filter {
```
bachand, line 186: Nit: naming

bachand (approved):
This is very powerful.

Sources/SwiftInspectorVisitors/StructVisitor.swift
dfed: I do want to consolidate at some point in the future. At a minimum we have good tests so we should be catch when we screw this up. I've been focused on getting something working this last week rather than generalizing the reused code – there's certainly room to go back and make this code less repetitive. Not sure how to do that just yet though.

Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
dfed: e8531b59c12a28b8f76fd8f7afc1b85b76a3e9e2

Sources/SwiftInspectorVisitors/FileVisitor.swift
dfed: Because we can only parse one extension at a time since extensions are top-level only.

To illustrate, the following code compiles:
```
struct Foo {
  struct Bar {}
}
```

And the following code doesn't:
```
extension Foo {
  extension Bar {}
}
```

Sources/SwiftInspectorVisitors/Tests/ClassVisitorSpec.swift
dfed: I made the `it` fully qualified in other tests. I'll do the same here.

265799912675f0555a89354a524afb27834bbe9e

Sources/SwiftInspectorVisitors/Tests/FileVisitorSpec.swift
dfed: yeah I did that elsewhere. Good call on doing it everywhere.

ecbf472f8ca10b6d4e402bee535af689182d8071

Sources/SwiftInspectorVisitors/StructVisitor.swift
bachand: Seems like we are doing the right thing for now.

PR #68, opened 2021-01-30
Handle qualified names when parsing types

dfed:
I realized this morning that we weren't going to properly parse `Foundation.Array` as a type. This PR creates an extension that aids with the extraction of names from a `TypeSyntax` object, and utilizes it in the places where we were previously either inspecting the `description` directly or force-casting to a `SimpleTypeIdentifierSyntax`.

### DIFF
--- a/Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
@@ -71,6 +71,44 @@ final class GenericRequirementVisitorSpec: QuickSpec {
         }
       }
 
+      context("visiting a syntax tree involving a protocol that has a left-side fully qualified generic constraint") {
+        it("finds the generic requirements") {
+          let content = """
+            public protocol SomeGenericProtocol: GenericProtocol where
+              FooModule.LeftType == RightType
+            {}
+            """
+
+          try VisitorExecutor.walkVisitor(
+            self.sut,
+            overContent: content)
+
+          expect(self.sut.genericRequirements.first) == GenericRequirement(
+            leftType: "FooModule.LeftType",
+            rightType: "RightType",
+            relationship: .equals)
+        }
+      }
+
+      context("visiting a syntax tree involving a protocol that has a right-side fully qualified generic constraint") {
+        it("finds the generic requirements") {
+          let content = """
+            public protocol SomeGenericProtocol: GenericProtocol where
+              LeftType == FooModule.RightType
+            {}
+            """
+
+          try VisitorExecutor.walkVisitor(
+            self.sut,
+            overContent: content)
+
+          expect(self.sut.genericRequirements.first) == GenericRequirement(
+            leftType: "LeftType",
+            rightType: "FooModule.RightType",
+            relationship: .equals)
+        }
+      }
+
       context("visiting a syntax tree involving a protocol that has two generic constraints") {
         beforeEach {
           let content = """
--- a/Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
@@ -1,0 +1,111 @@
+// Created by Dan Federman on 2/1/21.
+//
+// Copyright © 2021 Dan Federman
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import Nimble
+import Quick
+import SwiftInspectorTestHelpers
+import SwiftSyntax
+
+@testable import SwiftInspectorVisitors
+
+final class TypeSyntaxNameExtractorSpec: QuickSpec {
+  private var sut = EnumVisitor()
+
+  override func spec() {
+    beforeEach {
+      self.sut = EnumVisitor()
+    }
+
+    describe("qualifiedName") {
+      context("when called on a TypeSyntax node representing a SimpleTypeIdentifierSyntax") {
+        final class SimpleTypeIdentifierSyntaxVisitor: SyntaxVisitor {
+          var simpleTypeIdentifier: String?
+          override func visit(_ node: SimpleTypeIdentifierSyntax) -> SyntaxVisitorContinueKind {
+            simpleTypeIdentifier = TypeSyntax(node).qualifiedName
+            return .skipChildren
+          }
+        }
+
+        var visitor: SimpleTypeIdentifierSyntaxVisitor!
+        beforeEach {
+          let content = """
+              var int: Int = 1
+              """
+
+          visitor = SimpleTypeIdentifierSyntaxVisitor()
+          try? VisitorExecutor.walkVisitor(visitor, overContent: content)
+        }
+
+        it("Finds the type") {
+          expect(visitor?.simpleTypeIdentifier) == "Int"
+        }
+      }
+
+      context("when called on a TypeSyntax node representing a MemberTypeIdentifierSyntax") {
+        final class MemberTypeIdentifierSyntaxVisitor: SyntaxVisitor {
+          var memberTypeIdentifier: String?
+          override func visit(_ node: MemberTypeIdentifierSyntax) -> SyntaxVisitorContinueKind {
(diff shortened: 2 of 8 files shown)


Sources/SwiftInspectorVisitors/TypeInheritanceVisitor.swift
```diff
@@ -30,7 +30,7 @@ public final class TypeInheritanceVisitor: SyntaxVisitor {
   public private(set) var inheritsFromTypes = [String]()
 
   public override func visit(_ node: InheritedTypeSyntax) -> SyntaxVisitorContinueKind {
-    inheritsFromTypes.append(node.typeName.description.trimmingCharacters(in: .whitespacesAndNewlines))
+    inheritsFromTypes.append(node.typeName.qualifiedName)
```
dfed, line 33: These two lines are functionally equivalent

Sources/SwiftInspectorVisitors/TypeSyntax+NameExtractor.swift
```diff
@@ -0,0 +1,25 @@
...
+  /// Returns the the qualified name for the type the receiver represents.
+  var qualifiedName: String {
+    if let typeIdentifier = self.as(SimpleTypeIdentifierSyntax.self) {
+      return typeIdentifier.name.text
+
+    } else if let typeIdentifier = self.as(MemberTypeIdentifierSyntax.self) {
+      let baseName = typeIdentifier.baseType.qualifiedName
+      return "\(baseName).\(typeIdentifier.name.text)"
+
+    } else {
```
dfed, line 17: I don't know how to test this `else` case since it's explicitly covering a case that isn't expected. Even if I were convinced that both `SimpleTypeIdentifierSyntax` and `MemberTypeIdentifierSyntax` were the only types that we could encounter, there's no way for me to avoid an `else` without a force unwrap somewhere.

Given that I don't know how to test this condition, I think we're likely going to need to merge through the code coverage reduction. Open to ideas on how to avoid doing that.

bachand: @dfed I think we could test this by manually constructing a `TypeSyntax` object with an invalid `raw.kind` value. If the value is anything other than `.simpleTypeIdentifier` or `.memberTypeIdentifier` we will end up in the `else`.

https://github.com/apple/swift-syntax/blob/577d3bbe3a4f39ba8843093f3bd1ae0b593da557/Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxTypeNodes.swift#L68

https://github.com/apple/swift-syntax/blob/577d3bbe3a4f39ba8843093f3bd1ae0b593da557/Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxTypeNodes.swift#L175

bachand: The problem will then run into is that we're going to assert, but we could fix that too by having an internal function we call into from here.

```swift
extension TypeSyntax {

  var qualifiedName: String { computeQualifiedName(assertionFailure) }

  func computeQualifiedName(_ failureClosure: (String) -> Void) { /* ... */ }
}
```

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -71,6 +71,44 @@ final class GenericRequirementVisitorSpec: QuickSpec {
         }
       }
 
+      context("visiting a syntax tree involving a protocol that has a left-side fully qualified generic constraint") {
+        it("finds the generic requirements") {
+          let content = """
+            public protocol SomeGenericProtocol: GenericProtocol where
+              FooModule.LeftType == RightType
```
bachand, line 78: Would this code be value? I thought that the left type needed to be on the protocol itself? Seeing an example that compiles would help me.

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -71,6 +71,44 @@ final class GenericRequirementVisitorSpec: QuickSpec {
...
+            public protocol SomeGenericProtocol: GenericProtocol where
+              LeftType == FooModule.RightType
+            {}
+            """
+
+          try VisitorExecutor.walkVisitor(
+            self.sut,
+            overContent: content)
+
+          expect(self.sut.genericRequirements.first) == GenericRequirement(
```
bachand, line 105: 👌

Sources/SwiftInspectorVisitors/TypeInheritanceVisitor.swift
bachand: To clarify though... we are no longer stripping whitespace. But that is OK since we're pulling the name from a more robust place now?

bachand (approved):

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
dfed: yeah I thought that too! Then I put the following code into a playground and it compiled:
```
extension Array where
  AnyObject == Element
{}
```

Not that I think we should write code like this, but we should probably support someone doing so.

dfed: And since that example used extensions rather than protocols, check out this code which also compiles:
```
protocol GenericTest {
  associatedtype SomeType
}

protocol GenericWhereTest: GenericTest where AnyObject == SomeType {}
```

Sources/SwiftInspectorVisitors/TypeInheritanceVisitor.swift
dfed: yup! and our unit tests prove that out 🙂

Sources/SwiftInspectorVisitors/TypeSyntax+NameExtractor.swift
dfed: Hadn't occurred to me that we'd be able to create an unexpected object that conformed to `TypeSyntax`. Managed to get it done in 744be8c5e6e25d7600504250cac3d8560b6c82bd

Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
```diff
@@ -0,0 +1,111 @@
...
+          }
+        }
+
+        it("asserts") {
+          expect(try VisitorExecutor.walkVisitor(
+                  ArraySyntaxVisitor(),
+                  overContent: """
+                    var intArray: [Int] = [Int]()
+                    """))
+            .to(throwAssertion())
```
bachand, line 106: I didn't know you could catch assertions. Nice!

Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
```diff
@@ -0,0 +1,111 @@
...
+          expect(visitor?.memberTypeIdentifier) == "Swift.Int"
+        }
+      }
+
+      context("when called on a TypeSyntax node representing an Array") {
+        final class ArraySyntaxVisitor: SyntaxVisitor {
+          override func visit(_ node: ArrayTypeSyntax) -> SyntaxVisitorContinueKind {
+            // This line should assert.
+            _ = TypeSyntax(node).qualifiedName
+            fail("This line should never be reached")
```
bachand, line 95: Is this function from Quick/Nimble?

Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
dfed: It's the one thing that makes me happy to work with Nimble tests.

Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
dfed: yup! it's the nimble equivalent of `XCTFail`

Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
```diff
@@ -0,0 +1,111 @@
...
+            return .skipChildren
+          }
+        }
+
+        it("asserts") {
+          expect(try VisitorExecutor.walkVisitor(
+                  ArraySyntaxVisitor(),
+                  overContent: """
+                    var intArray: [Int] = [Int]()
```
bachand, line 104: One random thought... Functionally `[Int]` is equivalent to `Array<Int>`. I wonder if we need to handle this case when we think about properties. I would expect that most users of SwiftInspector may not care if the code says `[Int]` or `Array<Int>`.

Sources/SwiftInspectorVisitors/TypeSyntax+NameExtractor.swift
```diff
@@ -0,0 +1,25 @@
+// Created by dan_federman on 1/30/21.
+// Copyright © 2021 Airbnb Inc. All rights reserved.
+
+import SwiftSyntax
+
+extension TypeSyntax {
+
+  /// Returns the the qualified name for the type the receiver represents.
+  var qualifiedName: String {
```
bachand, line 9: Let's clarify in the documentation when it's safe to call this.

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
bachand: Good to know.

bachand (approved):

Sources/SwiftInspectorVisitors/TypeSyntax+NameExtractor.swift
dfed: good call. I may eventually want to factor this out to make it more compile-time safe.

2fcfd31a00eff9d58c21f9748d27fab349d8bb97

Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
dfed: Indeed. The good news is that our current code doesn't yet inspect an `ArrayTypeSyntax`, and the `qualifiedName` property is `internal` so if we end up tripping the assertion it's because we've foot-gunned.

But yes, there's room to make this code more compile-time safe.

bachand: And in addition to compile-time safe I am hoping we can make the output as easy as possible to consume. In most cases I suspect you won't care if a type is `[Int]` or `Array<Int>`.

PR #69, opened 2021-02-03
Introduce TypeDescription

dfed:
It turns out handling parsing qualified names (introduced in #68) is only the beginning of what's required to properly parse out types in a complex codebase.

This PR creates and implements utilizing a `TypeDescription`, which is an indirect `enum` that is used to encode explicit type information. We adopt `TypeDescription` in this PR and stop using `String` to represent types in much of the code.

This PR is smaller than it looks. Lots of tests had to be updated in a rote manner once I introduced this type and moved away from utilizing `String` to represent types. I've highlighted with comments where the meat of this PR lies.

Note that this PR does not meet codecov requirements – see comments in the PR. I'm curious to hear @fdiaz's thoughts on whether this is problematic.


Sources/SwiftInspectorVisitors/ClassVisitor.swift
```diff
@@ -27,8 +27,8 @@ import SwiftSyntax
 
 public final class ClassVisitor: SyntaxVisitor {
 
-  public init(parentTypeName: String? = nil) {
-    self.parentTypeName = parentTypeName
+  public init(parentType: TypeDescription? = nil) {
```
dfed, line 30: `parentTypeName` became `parentType` everywhere.

Sources/SwiftInspectorVisitors/ClassVisitor.swift
```diff
@@ -52,11 +52,8 @@ public final class ClassVisitor: SyntaxVisitor {
       // Base case. We've previously found a class declaration, so this must be an inner class.
       // This class visitor shouldn't recurse down into the children.
       // Instead, we'll use a new class visitor to get the information from this class.
-      let qualifiedParentTypeName = QualifiedParentNameCreator.createNameGiven(
```
dfed, line 55: `QualifiedParentNameCreator` was deleted in favor of a `createTypeWithName(_:parent:)` method on `TypeDescription`. This change was made everywhere.

Sources/SwiftInspectorVisitors/Tests/ClassVisitorSpec.swift
```diff
@@ -51,8 +51,8 @@ final class ClassVisitorSpec: QuickSpec {
 
             let classInfo = self.sut.classes.first
             expect(classInfo?.name) == "SomeClass"
-            expect(classInfo?.inheritsFromTypes) == []
-            expect(classInfo?.parentTypeName).to(beNil())
+            expect(classInfo?.inheritsFromTypes.map { $0.description }) == []
```
dfed, line 54: In order to keep the test signatures the same (and also test `description`), I always inspect the `description` at the test sites. It isn't necessary on this line, but it's necessary in most spots. See below for how it's used, but I don't think these test files need detailed review (except where called out explicitly).

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,346 @@
...
+
+import Foundation
+import Nimble
+import Quick
+import SwiftInspectorTestHelpers
+import SwiftSyntax
+
+@testable import SwiftInspectorVisitors
+
+final class TypeDescriptionSpec: QuickSpec {
```
dfed, line 33: This file should be reviewed in detail

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,346 @@
...
+        }
+
+        it("throws") {
+          expect(try decoder.decode(TypeDescription.self, from: data)).to(throwError())
+        }
+      }
+    }
+
+    describe("typeDescription") {
+      context("when called on a TypeSyntax node representing a SimpleTypeIdentifierSyntax") {
```
dfed, line 142: The first 2 of these contexts were stolen from `TypeSyntax+NameExtractorSpec.swift `, which was delete din this PR. Everything else is pretty new.

Sources/SwiftInspectorVisitors/Tests/TypeSyntax+NameExtractorSpec.swift
```diff
@@ -1,111 +0,0 @@
...
-        final class ArraySyntaxVisitor: SyntaxVisitor {
-          override func visit(_ node: ArrayTypeSyntax) -> SyntaxVisitorContinueKind {
-            // This line should assert.
-            _ = TypeSyntax(node).qualifiedName
-            fail("This line should never be reached")
-            return .skipChildren
-          }
-        }
-
-        it("asserts") {
```
dfed, line 100: Note that I am no longer creating an assertion case. I can add an assertion case back... but eventually we're likely to need to handle all types. I'm not sure testing the `unknown` case is particularly helpful.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,231 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import SwiftSyntax
+
```
dfed, line 26: This file should be reviewed in detail

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,231 @@
...
+public enum TypeDescription: Codable, Equatable, CustomStringConvertible {
+  case simple(name: String)
+  indirect case member(name: String, baseType: TypeDescription)
+  indirect case composition([TypeDescription])
+  indirect case optional(TypeDescription)
+  indirect case implicitlyUnwrapped(TypeDescription)
+  indirect case array(TypeDescription)
+  indirect case dictionary(key: TypeDescription, value: TypeDescription)
+  indirect case tuple([TypeDescription])
+  case unknown(text: String)
```
dfed, line 36: Note that I'm still missing cases for the following types:
```
 ClassRestrictionTypeSyntax
 SomeTypeSyntax
 MetatypeTypeSyntax
 FunctionTypeSyntax
 AttributedTypeSyntax
 UnknownTypeSyntax
```

This PR is big enough. We can add these types later as needed.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,231 @@
...
+    case let .member(name, baseType):
+      return "\(baseType).\(name)"
+    case let .tuple(types):
+      return "(\(types.map { $0.description }.joined(separator: ", ")))"
+    case let .unknown(text):
+      return text
+    }
+  }
+
+  public init(from decoder: Decoder) throws {
```
dfed, line 62: implementing encoding/decoding absolutely was a particular kind of fun 😅

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,234 @@
...
+  private static let compositionDescription = "composition"
+  private static let optionalDescription = "optional"
+  private static let implicitlyUnwrappedDescription = "implicitlyUnwrapped"
+  private static let arrayDescription = "array"
+  private static let dictionaryDescription = "dictionary"
+  private static let tupleDescription = "tuple"
+  private static let unknownDescription = "unknown"
+}
+
+extension TypeSyntax {
```
dfed, line 192: This extension works similarly to the now-deleted extension in `TypeSyntax+NameExtractor.swift`, but has more cases.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,234 @@
...
+
+    } else if let typeIdentifier = self.as(DictionaryTypeSyntax.self) {
+      return .dictionary(
+        key: typeIdentifier.keyType.typeDescription,
+        value: typeIdentifier.valueType.typeDescription)
+
+    } else if let typeIdentifiers = self.as(TupleTypeSyntax.self) {
+      return .tuple(typeIdentifiers.elements.map { $0.type.typeDescription })
+
+    } else {
```
dfed, line 227: I currently do not have coverage on this `else` statement. See [this comment](https://github.com/fdiaz/SwiftInspector/pull/69/files#r569045995) for more info.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+      return .dictionary(
+        key: typeIdentifier.keyType.typeDescription,
+        value: typeIdentifier.valueType.typeDescription)
+
+    } else if let typeIdentifiers = self.as(TupleTypeSyntax.self) {
+      return .tuple(typeIdentifiers.elements.map { $0.type.typeDescription })
+
+    } else if self.as(ClassRestrictionTypeSyntax.self) != nil {
+      // A class restriction is the same as requiring inheriting from AnyObject.
+      return .simple(name: "AnyObject")
```
dfed, line 229: Note that I'm being tricky here. `protocol SomeObject: class` is functionally equivalent to `protocol SomeObject: AnyObject`, so I'm just pretending that this is the latter rather than creating a new enum case to handle this somewhat deprecated way to restrict a protocol to a reference type.

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
```diff
@@ -83,10 +82,9 @@ final class GenericRequirementVisitorSpec: QuickSpec {
             self.sut,
             overContent: content)
 
-          expect(self.sut.genericRequirements.first) == GenericRequirement(
```
bachand, line 86: Why is `GenericRequirement` no longer equatable?

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import SwiftSyntax
+
+public enum TypeDescription: Codable, Equatable, CustomStringConvertible {
+  case simple(name: String)
```
bachand, line 28: This is a pain but can we document these cases, especially this one?

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import SwiftSyntax
+
+public enum TypeDescription: Codable, Equatable, CustomStringConvertible {
```
bachand, line 27: Let's add a doc comment here too.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+  case simple(name: String)
+  indirect case member(name: String, baseType: TypeDescription)
+  indirect case composition([TypeDescription])
+  indirect case optional(TypeDescription)
+  indirect case implicitlyUnwrapped(TypeDescription)
```
bachand, line 32:
```suggestion
  indirect case implicitlyUnwrappedOptional(TypeDescription)
```
?

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: What do you think about adding code comment to indicate these intentional omissions?

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+  indirect case array(TypeDescription)
+  indirect case dictionary(key: TypeDescription, value: TypeDescription)
+  indirect case tuple([TypeDescription])
+  case unknown(text: String)
+
+  /// A description of this type that can be used for code-generation to represent this type.
+  public var description: String {
```
bachand, line 39: I've trained myself to avoid using `description` for core API use cases under the assumption that it's not guaranteed to be stable ([example](https://twitter.com/mattt/status/1174041151214538752?s=20)). What do you think about having a specific property like `text`, even if `description` has the same value? Then we can choose to evolve `description` if we want and not break use cases.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+
+  enum CodingError: Error {
+    case unknownCase
+  }
+
+  /// - Parameters:
+  ///   - name: The simple name of the returned type.
+  ///   - parent: The parent type for the returned type.
+  /// - Returns: Returns a type description of case `.member` with the given name as the name and the receiver as the base type.
+  /// - Note: This method only makes sense when the `parent` is of case  `simple`, `member`, `optional`, and `implicitlyUnwrapped`.
```
bachand, line 149: Is there a way we can enforce this?

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import SwiftSyntax
+
+public enum TypeDescription: Codable, Equatable, CustomStringConvertible {
+  case simple(name: String)
+  indirect case member(name: String, baseType: TypeDescription)
```
bachand, line 29: Member is a pretty specific concept in Swift. I've heard it refer to instance properties and functions of a type. I think you are using `.member` here when we have a nested type. What about `.nested`?

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+    case let .dictionary(key, value):
+      try container.encode(key, forKey: .typeDescriptionKey)
+      try container.encode(value, forKey: .typeDescriptionValue)
+    case let .member(name, baseType):
+      try container.encode(name, forKey: .text)
+      try container.encode(baseType, forKey: .typeDescription)
+    }
+  }
+
+  enum CodingKeys: String, CodingKey {
```
bachand, line 132: Documenting what these are intended to be used for would be helpful.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+    if caseDescription == Self.simpleDescription {
+      let text = try values.decode(String.self, forKey: .text)
+      self = .simple(name: text)
+
+    } else if caseDescription == Self.unknownDescription {
+      let text = try values.decode(String.self, forKey: .text)
+      self = .unknown(text: text)
+
+    } else if caseDescription == Self.memberDescription {
+      let text = try values.decode(String.self, forKey: .text)
```
bachand, line 74: I'm having trouble following why we use `.text` in some places and `.textDescription` in others.

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: For the code review gods, can we link to some source showing that they are equivalent?

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: I see some value in verifying the assertion so we know that we have that guardrail in place if the language expands over time and we no longer cover all cases.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,238 @@
...
+  enum CodingError: Error {
+    case unknownCase
+  }
+
+  /// - Parameters:
+  ///   - name: The simple name of the returned type.
+  ///   - parent: The parent type for the returned type.
+  /// - Returns: Returns a type description of case `.member` with the given name as the name and the receiver as the base type.
+  /// - Note: This method only makes sense when the `parent` is of case  `simple`, `member`, `optional`, and `implicitlyUnwrapped`.
+  static func createTypeWithName(_ name: String, parent: TypeDescription?) -> TypeDescription {
```
bachand, line 150:
```suggestion
  static func typeDescriptionWithName(_ name: String, parent: TypeDescription?) -> TypeDescription {
```
?

Sources/SwiftInspectorVisitors/Tests/TypeInheritanceVisitorSpec.swift
```diff
@@ -88,7 +88,21 @@ final class TypeInheritanceVisitorSpec: QuickSpec {
           }
 
           it("returns the conforming type names") {
-            expect(self.sut.inheritsFromTypes) == ["Foo", "Bar", "FooBar"]
+            expect(self.sut.inheritsFromTypes.map { $0.description }) == ["Foo", "Bar", "FooBar"]
+          }
+        }
+
+        context("with a composition conformances on multiple lines") {
```
bachand, line 95:
```suggestion
        context("with composition conformances on multiple lines") {
```

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,383 @@
...
+    describe("init(from decoder: Decoder)") {
+      let encoder = JSONEncoder()
+      let decoder = JSONDecoder()
+      var data: Data!
+      var sut: TypeDescription!
+
+      context("when decoding a simple type") {
+        beforeEach {
+          sut = .simple(name: "Foo")
+          data = try? encoder.encode(sut)
```
bachand, line 45: I am curious what @fdiaz thinks but I feel like these tests would be more powerful if we wrote out the JSON for the decoding tests rather than creating it by the encode method (and vice versa). That way we guard against symmetrical bugs in our encode/decode logic.

bachand: I'm going to hold off on reviewing this file in detail right now since I feel like it would be good to align on this first.

bachand (commented):
I really like the direction here @dfed .

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,383 @@
...
+          sut = .dictionary(key: .simple(name: "Foo"), value: .simple(name: "Bar"))
+          data = try? encoder.encode(sut)
+        }
+
+        it("to decode the encoded type description") {
+          expect(try decoder.decode(TypeDescription.self, from: data)) == sut
+        }
+      }
+
+      context("when decoding a dictionary type") {
```
dfed, line 108:
```suggestion
      context("when decoding a composition type") {
```

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,383 @@
...
+        beforeEach {
+          sut = .composition([.simple(name: "Foo"), .optional(.simple(name: "Bar"))])
+          data = try? encoder.encode(sut)
+        }
+
+        it("to decode the encoded compositionDescription description") {
```
dfed, line 114:
```suggestion
        it("to decode the encoded composition description") {
```

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I love it.

dd148f5f47478e951c361e750081533bb9e6b81d

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Yeah the tough thing is I'm running low on ways to create this failing test 😅

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: https://forums.swift.org/t/class-only-protocols-class-vs-anyobject/11507/4

Sources/SwiftInspectorVisitors/ClassVisitor.swift
```diff
@@ -52,11 +52,8 @@ public final class ClassVisitor: SyntaxVisitor {
...
-        currentParentTypeName: parentTypeName,
-        currentTypeName: classInfo.name)
-
-      let innerClassVisitor = ClassVisitor(parentTypeName: qualifiedParentTypeName)
+      let newParentType = TypeDescription.createTypeWithName(classInfo.name, parent: self.parentType)
```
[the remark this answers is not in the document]
dfed: good catch! 5c91af39c1016905e9a9833bd844ff56cec45494

Sources/SwiftInspectorVisitors/ClassVisitor.swift
```diff
@@ -137,15 +130,15 @@ public final class ClassVisitor: SyntaxVisitor {
 
   // MARK: Private
 
-  private let parentTypeName: String?
+  private let parentType: TypeDescription?
```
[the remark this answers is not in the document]
dfed: Because in the following sample, `InnerClass` has a parent type of `OuterClass`. This helps us create a fully-qualified name `OuterClass.InnerClass`.
```
final class OuterClass {
  final class InnerClass {}
}
```

Note that the outer type doesn't _need_ to be a class. It could be an enum or struct.

Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
```diff
@@ -124,8 +123,8 @@ public final class ExtensionVisitor: SyntaxVisitor {
 }
 
 public struct ExtensionInfo: Codable, Equatable {
-  public let name: String
-  public private(set) var inheritsFromTypes: [String]
+  public let type: TypeDescription
```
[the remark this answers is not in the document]
dfed: since this is a lower-case property I think we're okay as-is for now. At least we're consistent with the property naming (I think) 😬

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
dfed: it is! But I'm utilizing feedback I got from @fdiaz in another PR to only check the fields we care about, so we don't need to update each of these tests if we ever add a new property to `GenericRequirement`.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: 471449b5c729c8fe09f1b63133e7b3241fefaa40

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I'm following `SwiftSyntax`'s lead here. Nodes all have a `description` that is source-accurate. While counter-intuitive, I'm trying to make this layer match the intent.

I think in the future if we wanted to add more info we'd need to evolve `debugDescription` rather than `description`, which seems okay? Open to changing here but wanted to explain my thinking before making the change blindly.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: We could `switch` and then assert internally... but I'm not sure that's worth it just yet. No way to compile-time enforce afaik.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I like diverging a bit from how Swift works internally to make the API easier. Also changing `baseType` to `parentType`.

56de1fa628f88bda61de9bbbb7dd9ed53bd69a81

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: @bachand I don't think I ever use `textDescription`. Did you mean something else?

dfed: ohhh `typeDescription`! Yeah I'll document the coding keys and that should make it more clear.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: dig it f7f701caf31319d28f8f82a475a86285e6e1c259

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: b84bcfb953dfce5e0cd0592b042d9976d914fdaa

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: 6a2a8769b0f364cfc075e3378452abbf76c4e47b

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: https://github.com/fdiaz/SwiftInspector/pull/69#discussion_r569709703

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: feb7816343f68dc68daba214e029afe86eae2a90

Hopefully the `description` also helps explain, but yes comments are good 🙂

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: feb7816343f68dc68daba214e029afe86eae2a90

dfed:
I have addressed all of @bachand's feedback, with the exception of https://github.com/fdiaz/SwiftInspector/pull/69#discussion_r569648323

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: also e4b38ac and c59a340 😅

Sources/SwiftInspectorVisitors/Tests/GenericRequirementVisitorSpec.swift
fdiaz: Yeah, I've found that it's pretty annoying to have to modify every single test that does this when you add a new property so I've used 2 different patterns:

1) What Dan has here
2) Create a thing factory method in tests with defaults values `GenericRequirement.fake(leftType: String = "", rightType: String ="", ...` so when we add a new property we can just add it to this method instead with a default value and every test continues working

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
fdiaz: I agree with Michael here although we don't need a JSON, we can pass Data. So we could simply convert from a String to Data using `.data(using: .utf8)` and then pass it to a JSONDecoder / Encoder.

As is, this is too tied to the implementation of this type IMHO so these tests are not as great as they could be.

fdiaz (commented):
Mostly waiting for naming convention to be updated (type / description) and having more robust tests in TypeDescriptionSpec. Everything else LGTM!

Sources/SwiftInspectorVisitors/ClassVisitor.swift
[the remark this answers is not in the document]
bachand: Hmm I'm not sure I follow. Maybe to be more clear, what is the difference between `self.parentType` and `self.classInfo?.parentType`? An example may help get us on the same page.

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: Ah I didn't realize that you were basing "member" on SwiftSyntax types. Yeah I think we can diverge if it's more clear.

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: Very helpful!

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: Ah yes, `typeDescription`. I'm glad you figured it out.

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: I will defer to @fdiaz on whether meeting the code coverage requirement is a dealbreaker.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -173,8 +173,8 @@ public enum TypeDescription: Codable, Equatable, CustomStringConvertible {
   /// - Parameters:
   ///   - name: The simple name of the returned type.
   ///   - parent: The parent type for the returned type.
-  /// - Returns: Returns a type description of case `.member` with the given name as the name and the receiver as the base type.
-  /// - Note: This method only makes sense when the `parent` is of case  `simple`, `member`, `optional`, and `implicitlyUnwrappedOptional`.
+  /// - Returns: Returns a type description of case `.nested` with the given name as the name and the receiver as the base type.
```
bachand, line 176: We return `.simple` in some cases. Should we update this comment?

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,267 @@
...
+    case typeDescription
+    /// The value for this key is the associated value of type [TypeDescription]
+    case typeDescriptions
+    /// The value for this key is the associated value of type TypeDescription that represents the key in a dictionary.
+    case typeDescriptionDictionaryKey
+    /// The value for this key is the associated value of type TypeDescription that represents the value in a dictionary
+    case typeDescriptionDictionaryValue
+  }
+
+  enum CodingError: Error {
```
bachand, line 169: Should we make this public so someone could identify it in their consuming code?

bachand (commented):
It seems like there are a few more comments to address given @fdiaz 's review. Really like the adjustments so far.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,267 @@
...
+    case unknownCase
+  }
+
+  /// - Parameters:
+  ///   - name: The simple name of the returned type.
+  ///   - parent: The parent type for the returned type.
+  /// - Returns: Returns a type description of case `.nested` with the given name as the name and the receiver as the base type.
+  /// - Note: This method only makes sense when the `parent` is of case  `simple`, `nested`, `optional`, and `implicitlyUnwrappedOptional`.
+  static func typeDescriptionWithName(_ name: String, parent: TypeDescription?) -> TypeDescription {
```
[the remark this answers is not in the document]
dfed: that works too!

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Alright let's see what y'all think of this naming idea: 167f0172d5e21c5d0b1b7bb9cd49f4ceb5d4716e

Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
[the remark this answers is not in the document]
dfed: 63c5a84d102f376126124302dcb708509dd2ee01

Sources/SwiftInspectorVisitors/ClassVisitor.swift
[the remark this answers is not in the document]
dfed: Aha! I'm pretty sure they end up being the same. But we don't have a `self.classInfo` at `init`, and I needed somewhere to store the `parentType` until we had one.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: 15e101be77005b904a638417a1c91d26638d877b

Sources/SwiftInspectorVisitors/TypeDescription.swift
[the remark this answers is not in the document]
dfed: 15e101be77005b904a638417a1c91d26638d877b

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,267 @@
...
+   * MetatypeTypeSyntax
+   * FunctionTypeSyntax
+   * AttributedTypeSyntax
+   * UnknownTypeSyntax
+   *
+   * We will likely need to add these types at some point in the future.
+   * We’ll Get There™
+   */
+
+  /// A description of this type that can be used for code-generation to represent this type.
```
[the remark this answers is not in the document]
dfed: with the latest update I'm less sure that's required. lmk what you think.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -0,0 +1,267 @@
...
+      return "(\(types.map { $0.description }.joined(separator: ", ")))"
+    case let .unknown(text):
+      return text
+    }
+  }
+
+  public init(from decoder: Decoder) throws {
+    let values = try decoder.container(keyedBy: CodingKeys.self)
+    let caseDescription = try values.decode(String.self, forKey: .caseDescription)
+    if caseDescription == Self.simpleDescription {
```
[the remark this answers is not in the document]
dfed: To avoid creating more conflicts for myself I'll waive off this until a later date since this feedback is purely style and doesn't affect API or function.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: 9c375bad2a837b3f8a1fa027828df3b68c84523f

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
dfed: so, I think I've done what y'all suggested in b9bb7ed151f4ff3d3ab3e823255f6f47d163809a. But I truly do not understand how this is more likely to catch a failure or is more maintainable vs what I had.

dfed:
@fdiaz @bachand all feedback has been addressed. PTAL.

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
dfed: I'm iterating on top of this in another branch to better handle generics. This branch requires changing the shape of this enum somewhat. This change has made it _much_ harder to ship that change. I am very curious what the benefit of this change is 😅

dfed: Concrete example of the types of changes are [here](https://github.com/fdiaz/SwiftInspector/pull/71/files#diff-e905411ed579d068d03cacb0e10f43774cdfdee6b3bcb046630b740d0866a6ea)

Sources/SwiftInspectorVisitors/TypeDescription.swift
fdiaz: I think it's fine, it's an edge case so I'm ok merging through the failing CI check (if that's possible, I'm not sure). If not, we can lower the allowed number per PR to 0.3%

Sources/SwiftInspectorVisitors/TypeDescription.swift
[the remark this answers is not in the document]
fdiaz: @dfed I think it'd be nice to have this documentation. One of the main things about SwiftSyntax is that whatever you _read_ you can _write_ and it will output the same thing always see [this](https://github.com/apple/swift/blob/main/lib/Syntax/README.md#design-and-implementation-guidelines)

> Represent Swift source with "full fidelity" - parsing a source file and printing the syntax tree should result in the same file.

We're deviating from this by not representing the exact notation of the source code which is totally fine for our use case but we're also being opinionated on style here which is something we should at least put in a comment IMHO

fdiaz (approved):
Thanks for addressing feedback! Some minor comments but this LGTM. I'd love a final look from @bachand before we merge though if possible.

Sources/SwiftInspectorVisitors/TypeDescription.swift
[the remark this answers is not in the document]
fdiaz: Particularly in this https://github.com/fdiaz/SwiftInspector/pull/69/files#diff-6a8854acd551d0fde3152864ddc0e807bae90c7b740b33078bffb630c1798356R76

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,497 @@
...
+
+        it("to decode the encoded type description") {
+          expect(try encoder.encode(sut)) == data
+        }
+      }
+
+      context("when encoding a composition type") {
+        beforeEach {
+          sut = .composition([.simple(name: "Foo"), .optional(.simple(name: "Bar"))])
+          data = "{\"caseDescription\":\"composition\",\"typeDescriptions\":[{\"caseDescription\":\"simple\",\"text\":\"Foo\"},{\"caseDescription\":\"optional\",\"typeDescription\":{\"caseDescription\":\"simple\",\"text\":\"Bar\"}}]}".data(using: .utf8)
```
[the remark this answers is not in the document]
dfed: making them multiline would make them even harder to maintain 😬. Note that when viewing this source in Xcode it's easier to read – not great, but easier.

I'd love to have more discussion on https://github.com/fdiaz/SwiftInspector/pull/69#discussion_r569835363 to better elucidate this format is preferred.

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,497 @@
...
+      var sut: TypeDescription!
+
+      context("when decoding a simple type") {
+        beforeEach {
+          sut = .simple(name: "Foo")
+          data = "{\"caseDescription\":\"simple\",\"text\":\"Foo\"}".data(using: .utf8)
+        }
+
+        it("to decode the encoded type description") {
+          expect(try decoder.decode(TypeDescription.self, from: data)) == sut
```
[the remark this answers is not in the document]
dfed: that makes a bunch of sense. 6601b190bd75ddbc6a30e05c53441f2eba283965

Sources/SwiftInspectorVisitors/TypeDescription.swift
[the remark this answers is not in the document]
dfed: Okay. Let me better document this in https://github.com/fdiaz/SwiftInspector/pull/71 since then I can say that the output is a canonical representation (which indicates that it isn't necessarily the same as the original)

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
bachand: From my point of view, the benefits are:
- Writing out the strings prevents against symmetrical bugs. So for example, let's say that you updated `encode(...)` and `decode(...)` to both interact with the wrong key. The old version of this test would pass. The new version wouldn't since the expected data is decoupled from the implementation.
- The new version of the tests are better documenting. I can read the tests and I learn more about how this type works. In the previous world I couldn't do that.

From the perspective of a reviewer of future changes, it can be nice to see how code changes affect real encoded JSON, which we now get. But I see how it's a pain from an authorship POV 😓 

Curious to get @fdiaz 's thoughts as he is deeper in testing best practices than I am.

bachand (approved):
Good work @dfed . This was a tricky one.

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
dfed: > Writing out the strings prevents against symmetrical bugs. So for example, let's say that you updated encode(...) and decode(...) to both interact with the wrong key. 

Pushing back on this a little bit: what does it mean for a key to be "wrong" in this context? We aren't interacting with a 3rd-party API where we need to abide by a strict contract – if we were, then I'd agree starting with JSON or Data is the right move. I'm pretty sure all we care about here is that we can encode & persist data and then decode that same data back into the original form.

> The new version of the tests are better documenting. I can read the tests and I learn more about how this type works. In the previous world I couldn't do that.

Entirely fair, though I'd wager the investment required to maintain these tests as-is isn't necessarily worth that win.

Also curious to hear Francisco's thoughts 🙂

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,487 @@
...
+    describe("init(from decoder: Decoder)") {
+      let decoder = JSONDecoder()
+      var data: Data!
+
+      context("when decoding a simple type") {
+        beforeEach {
+          data = "{\"caseDescription\":\"simple\",\"text\":\"Foo\"}".data(using: .utf8)
+        }
+
+        it("to decode the encoded type description") {
```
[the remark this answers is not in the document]
dfed: good catch! 35bda2315c36dc4728937109eda7e6090b097567

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,487 @@
...
+        }
+      }
+
+      context("when decoding an unknown case") {
+        beforeEach {
+          data = "{\"garbage\": \"fire\"}".data(using: .utf8)
+        }
+
+        it("throws") {
+          expect(try decoder.decode(TypeDescription.self, from: data)).to(throwError())
```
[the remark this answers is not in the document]
dfed: yes we should 🙂 af91054ba19430ce3eb35fe27d7fef84fcdc4720

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,487 @@
...
+      var data: Data!
+      var sut: TypeDescription!
+
+      context("when encoding a simple type") {
+        beforeEach {
+          sut = .simple(name: "Foo")
+          data = "{\"caseDescription\":\"simple\",\"text\":\"Foo\"}".data(using: .utf8)
+        }
+
+        it("to decode the encoded type description") {
```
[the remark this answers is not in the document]
dfed: copy/paste 🤦‍♂️ 975166d9ed0bbc0a398f79682883baa9e147fa83

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,487 @@
...
+          expect(try encoder.encode(sut)) == data
+        }
+      }
+
+      context("when encoding a nested type") {
+        beforeEach {
+          sut = .nested(name: "Bar", parentType: .simple(name: "Foo"))
+          data = "{\"caseDescription\":\"nested\",\"text\":\"Bar\",\"typeDescription\":{\"caseDescription\":\"simple\",\"text\":\"Foo\"}}".data(using: .utf8)
```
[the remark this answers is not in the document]
dfed: Good point. To reduce conflicts I'll clean this up on the next PR though.

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,487 @@
...
+
+        it("Finds the type") {
+          expect(visitor?.nestedTypeIdentifier?.asSource) == "Swift.Int"
+        }
+      }
+
+      context("when called on a TypeSyntax node representing a CompositionTypeSyntax") {
+        final class CompositionTypeSyntaxVisitor: SyntaxVisitor {
+          var composedTypeIdentifier: TypeDescription?
+          // Note: ideally we'd visit a node of type CompositionTypeElementListSyntax
```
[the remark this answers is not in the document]
dfed: `CompositionTypeElementListSyntax` is a list of composed types, where a list of composed types looks like `Foo & Bar`.

`InheritedTypeSyntax` tells us what type(s) a type declaration inherits from. So in this case, the inherited type of `protocol FooBar` is `Foo & Bar`. And since `InheritedTypeSyntax` has a `TypeSyntax` property, we can easily call `typeDescription` to determine the underlying type description.

Lists on the other hand don't have a way to access the types in the list without utilizing a `SyntaxVisitor`. Hence the use of `InheritedTypeSyntax` and the corresponding note 🙂

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -0,0 +1,487 @@
...
+            var int: Int!
+            """
+
+          visitor = ImplicitlyUnwrappedOptionalTypeSyntaxVisitor()
+          try? VisitorExecutor.walkVisitor(visitor, overContent: content)
+        }
+
+        it("Finds the type") {
+          expect(visitor?.implictlyUnwrappedOptionalTypeIdentifier?.asSource)
+            == "Int!"
```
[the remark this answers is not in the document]
dfed: fc488a78ee4f8ffb8b8eecb3662d986d4f016f3a

dfed:
Merging through codecov failure.

Sources/SwiftInspectorVisitors/TypeDescription.swift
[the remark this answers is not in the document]
dfed: Done in e3e854089058eeb9ec1cb921c1a898eefdbaf918 in https://github.com/fdiaz/SwiftInspector/pull/71/

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
dfed: We discussed offline. What we want are two sets of tests:

1. Tests that ensure that we haven't accidentally made a breaking change in our encoded formatting.
2. Tests that ensure that we can decode the same value we encoded.

I've addressed this feedback in a4a42c57df9737712d1209df71a36571b4012388 in #71

bachand: Thanks for iterating here @dfed. I wish we were able to land on the solution more quickly. I appreciate your patience.

dfed: And I appreciate 1-day turnaround for reviews on massive PRs. We're moving quickly but landing in a good place. Takes a village.

PR #74, opened 2021-02-06
Small naming improvements in `TypeDescription` spec

bachand:
When reviewing https://github.com/fdiaz/SwiftInspector/pull/73 I noticed a couple places we could be more consistent.


dfed (approved):
Thank you!

PR #75, opened 2021-02-09
Get modifier information from types

dfed:
This PR creates a `DeclarationModifierVisitor` which can be used for finding declaration modifiers (e.g. "public", "internal", "private", etc).


dfed:
Whoops. Opened this too early. Closing. Will re-open when this works 😅

PR #77, opened 2021-02-10
Use correct attribute for exporting an import in tests

dfed:
We were using `@_export import Foo` when the actual syntax is `@_exported import Foo`.

It doesn't make a difference to the tests, but it's nice when the test code would compile 🙂


fdiaz (approved):
Ha! Thanks!

PR #79, opened 2021-02-16
Make 'make develop' kill Xcode

dfed:
If we don't kill Xcode prior to running `make develop` if `SwiftInspector.xcodeproj` is already open, Xcode will complain that it needs to reload the file from disk. On Xcode 12.4, the project reloading will never complete and it will not be possible to run unit tests again until Xcode is killed.

This PR makes killing Xcode part of the `make develop` workflow. Note that we need to wait for a short moment before opening Xcode back up after killing it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ bindir ?= $(prefix)/bin
 .PHONY: develop
 develop:
 	swift package generate-xcodeproj --xcconfig-overrides settings.xcconfig
+	(killall -9 Xcode && sleep 0.25) || true # Kills Xcode and waits for the process to terminate if Xcode is open.
 	open SwiftInspector.xcodeproj 
 
 .PHONY: build


Makefile
```diff
@@ -4,6 +4,7 @@ bindir ?= $(prefix)/bin
 .PHONY: develop
 develop:
 	swift package generate-xcodeproj --xcconfig-overrides settings.xcconfig
+	(killall -9 Xcode && sleep 0.25) || true # Kills Xcode and waits for the process to terminate if Xcode is open.
```
fdiaz, line 7: Should we move this to _before_ we generate the project and remove the sleep?

dfed: I'd be down to kill Xcode first, but it also feels reasonable to not kill Xcode if we fail to generate the project?

fdiaz: Good point. I'm a bit hesitant about this change as is because this would kill any Xcode project someone might have open. Do you think there's:

1) An option to only kill the process that has the SwiftInspector Xcode project?
2) Use environment variables to have a way to opt-out of this?

dfed: Those Xcode projects would re-open, so at least there's that. But there's no way to only kill the part of the Xcode process that has opened/cached the xcodeproj.

Environment variables (or a new command?) to opt out seems like a possibility.

From a usage POV: every time I run this command and it _doesn't_ do this I end up a little sad 😅

fdiaz: Let's merge this as is in the meantime and we can figure out a better usage if it becomes a problem.

fdiaz (approved):

Makefile
dfed: Excited to get your feedback after using it a bit 😄

PR #82, opened 2021-02-18
Create TypealiasVisitor and associated spec

dfed:
Also integrates `typealias`-finding into `ProtocolVisitor`, `ExtensionVisitor`, `NestableTypeVisitor`, and `FileVisitor`.

### DIFF
--- a/Sources/SwiftInspectorVisitors/GenericParameterVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/GenericParameterVisitor.swift
@@ -38,6 +38,6 @@ public final class GenericParameterVisitor: SyntaxVisitor {
 }
 
 public struct GenericParameter: Codable, Equatable {
-  let name: String
-  let inheritsFrom: TypeDescription?
+  public let name: String
+  public let inheritsFrom: TypeDescription?
 }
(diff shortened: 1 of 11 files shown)


Sources/SwiftInspectorVisitors/GenericParameterVisitor.swift
```diff
@@ -0,0 +1,43 @@
...
+      .init(
+        name: node.name.text,
+        inheritsFrom: node.inheritedType?.typeDescription))
+    return .skipChildren
+  }
+}
+
+public struct GenericParameter: Codable, Equatable {
+  let name: String
+  let inheritsFrom: TypeDescription?
```
fdiaz, line 42: I think we want these properties to be public too?

fdiaz (approved):

Sources/SwiftInspectorVisitors/GenericParameterVisitor.swift
dfed: 🤦‍♂️ yes

dfed, line 42:
```suggestion
  public let name: String
  public let inheritsFrom: TypeDescription?
```

PR #78, opened 2021-02-16
Collapse ClassVisitor, StructVisitor, and EnumVisitor into a single NestableTypeVisitor

dfed:
Apologies in advance for the size of this PR 😬. I promise that most of the code is moved and that my inline comments will make it clear what is new vs what is recycled. This PR is best reviewed commit-by-commit – each commit compiles.

Developing the `ClassVisitor`, `StructVisitor`, and `EnumVisitor` involved _a lot_ of copying and pasting. We knew while they were being developed that there was room to reuse code, but the structure at the time eluded me.

What I realized looking back is that `ClassVisitor`, `StructVisitor`, and `EnumVisitor` all do exactly the same work: they parse out information from a declaration type that can be nested in another declaration type. The structure and execution of these classes were all the same, save for unintentional minor differences.

This PR creates a `NestableTypeVisitor` that is capable of parsing out these three declaration types. Doing so enabled us to reuse more than 500 lines of code we had previously copy-and-pasted.

PR #84, opened 2021-02-19
Do not miscategorize nested type

dfed:
In the last commit (b5bdd1aaa8ddc54387e5cac5f2fff26a961f79c6) of https://github.com/fdiaz/SwiftInspector/pull/78, I introduced a bug. Previously the compiler was enforcing that I did this right – afterwords with the typealias we had to be a bit more careful. This PR fixes that bug, and also introduces more tests to ensure this doesn't happen again.

### DIFF
--- a/Sources/SwiftInspectorVisitors/AssociatedtypeVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/AssociatedtypeVisitor.swift
@@ -52,7 +52,7 @@ public final class AssociatedtypeVisitor: SyntaxVisitor {
   }
 }
 
-public struct AssociatedtypeInfo: Codable, Equatable {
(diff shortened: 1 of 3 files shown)


fdiaz (approved):

Sources/SwiftInspectorVisitors/AssociatedtypeVisitor.swift
```diff
@@ -52,7 +52,7 @@ public final class AssociatedtypeVisitor: SyntaxVisitor {
   }
 }
 
-public struct AssociatedtypeInfo: Codable, Equatable {
+public struct AssociatedtypeInfo: Codable, Hashable {
```
dfed, line 55: this change fixes a build break on `main` that I introduced when my last PR merged without first rebasing on top of latest `main`.

PR #85, opened 2021-03-05
Create portable binary of SwiftInspector

fdiaz:
Thanks to @keith for writing [this](https://www.smileykeith.com/2021/03/03/editing-rpaths/) amazing blog post that helped create this portable binary using SwiftSyntax.

I also added documentation on how to release a new version using this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 prefix ?= /usr/local
 bindir ?= $(prefix)/bin
+xcode_path ?= $(shell xcode-select -p)
+xcode_toolchain ?= $(xcode_path)/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx
 
 .PHONY: develop
 develop:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -22,3 +24,14 @@ uninstall:
 .PHONY: clean
 clean:
 	rm -rf .build
+
+# Following https://www.smileykeith.com/2021/03/03/editing-rpaths/
+.PHONY: release
```
dfed, line 29: it's be really nifty if this command took a version tag string as an argument and could run `git tag` as well.

Though maybe we'd still want the ability to `make:release_build` without that functionality. In any case, we can add this later.

Makefile
```diff
@@ -1,5 +1,7 @@
 prefix ?= /usr/local
 bindir ?= $(prefix)/bin
+xcode_path ?= $(shell xcode-select -p)
```
dfed, line 3: til `shell`

dfed (approved):

Makefile
fdiaz: Yeah, ideally I'd love to do everything from the command line and also release through homebrew but one thing at a time 😁

PR #87, opened 2021-04-08
Create public `TypeSyntaxVisitor` and update properties command to use modern patterns

bachand:
**I apologize for the large PR :( I believe it should be easy to follow commit-by-commit. Many commits are very straightforward.**

The motivation for this PR was to update the `properties` command to include the type name of each property. When I started to dig in I realized that the `properties` command does not yet expose a public visitor. As a result, I refocused on making that change. I take the existing `TypeSyntaxVisitor` and make it `public`, and update it to the modern API where it has a property of the information it found rather than a callback closure.

While I was here I renamed the "PropertiesAnalyzer*" files and types to `Properties*` as that seems to match how we name other commands. I also removed the analyzer class in favor of an extension on `StandardAnalyzer`, which seems like the modern approach. The tests on the property analyzer were moved to the visitor.

**This PR should introduce no functional changes.** My intention is to modernize the architecture of this command so we can expand/improve it in future PRs, and so that other libraries can leverage this functionality in a public visitor.

cc: @thedrick who wrote this command initially I believe


Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
```diff
@@ -79,24 +79,23 @@ public final class TypeSyntaxVisitor: SyntaxVisitor {
...
 private final class PropertySyntaxVisitor: SyntaxVisitor {
 
-  init(typeName: String, onNodeVisit: @escaping (_ info: TypeProperties.PropertyData) -> Void) {
+  init(typeName: String) {
     self.typeName = typeName
-    self.onNodeVisit = onNodeVisit
   }
 
+  /// Information about each of the properties found on the type.
+  private(set) var propertiesData = [TypeProperties.PropertyData]()
```
bachand: Now we have a property rather than a callback. This seems to be the modern approach.

Sources/SwiftInspectorAnalyzers/PropertyAnalyzer.swift
```diff
@@ -37,15 +37,11 @@ public final class PropertyAnalyzer: Analyzer {
 
   /// Finds property information for the provided type in the Swift file
   /// - Parameter fileURL: The fileURL where the Swift file is located
-  public func analyze(fileURL: URL) throws -> TypeProperties? {
+  public func analyze(fileURL: URL) throws -> Set<PropertyData>? {
     let syntax: SourceFileSyntax = try cachedSyntaxTree.syntaxTree(for: fileURL)
-    var result: TypeProperties?
-    let visitor = TypeSyntaxVisitor(typeName: typeName) { [unowned self] typeProperties in
```
bachand, line 42: `Update type syntax visitor to modern approach` is a tricky commit. I do a number of things:
- I move the `merge(...)` method from `TypeProperties` to `TypeSyntaxVisitor`. It no longer throws.
- I write tests for the new `merge(...)` method.
- I update `TypeSyntaxVisitor` to the modern API of having a property of what it found rather than a callback.
- I remove `TypeProperties`.

After the commit the project builds and all tests pass if you comment out PropertyAnalyzerSpec.

Sources/SwiftInspectorAnalyzers/PropertyAnalyzer.swift
```diff
@@ -37,15 +37,11 @@ public final class PropertyAnalyzer: Analyzer {
 
   /// Finds property information for the provided type in the Swift file
   /// - Parameter fileURL: The fileURL where the Swift file is located
-  public func analyze(fileURL: URL) throws -> TypeProperties? {
+  public func analyze(fileURL: URL) throws -> Set<PropertyData>? {
```
bachand, line 40: The only additional information we had in `TypeProperties` was the type name. But we already know that since the `TypeSyntaxVisitor` is instantiated with a type name.

Sources/SwiftInspectorAnalyzers/PropertyAnalyzer.swift
bachand, line 42: `Update type syntax visitor to modern approach` is a tricky commit. I do a number of things:
- I move the `merge(...)` method from `TypeProperties` to `TypeSyntaxVisitor`.
- `merge(...)` no longer throws since the way that `TypeSyntaxVisitor` is defined we should only ever find property data about the type in question.
- I write tests for the new `merge(...)` method.
- I update `TypeSyntaxVisitor` to the modern API of having a property of what it found rather than a callback.
- I remove `TypeProperties`.

After the commit the project builds and all tests pass if you comment out PropertyAnalyzerSpec.

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
```diff
@@ -28,37 +28,39 @@ public final class TypeSyntaxVisitor: SyntaxVisitor {
...
   }
 
   // MARK: Public
 
+  /// Information about each of the properties found on the type. `nil` if the type is not found.
+  public private(set) var propertiesData: Set<PropertyData>?
+
   public override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
     if node.identifier.text == typeName {
-      processNode(node, withName: node.identifier.text, members: node.members.members)
```
bachand: Since we are only trying to look up a specific type and we are only invoking `processNode(...)` when we find that type, let's simplify.

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
```diff
@@ -193,11 +210,9 @@ private final class PropertySyntaxVisitor: SyntaxVisitor {
...
 }
 
-// MARK: - TypeProperties
-
-/// Information about a type as well as its associated properties
-public struct TypeProperties: Hashable {
+// MARK: - PropertyData
 
+public struct PropertyData: Hashable {
   public struct Modifier: Hashable, OptionSet {
```
bachand, line 216: `Modifiers` is now nested under `PropertyData`

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
```diff
@@ -233,51 +248,12 @@ public struct TypeProperties: Hashable {
...
-
-  /// The name of the type.
+  /// The name of the property
   public let name: String
-  /// The properties on this type
-  public let properties: Set<PropertyData>
-}
-
-extension TypeProperties {
-  struct MergeError: Error, CustomStringConvertible {
```
bachand: Assuming we have good test coverage that we don't surface properties for types that we're not looking for, I don't think we need this.

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
```diff
@@ -95,12 +95,9 @@ final class TypeSyntaxVisitorSpec: QuickSpec {
       }
 
       context("when there are no properties") {
-        beforeEach {
```
bachand, line 50: With `VisitorExecutor` we don't need these `beforeEach(...)` invocations.

bachand: `let content`s are promoted to the parent scope with no changes.

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
```diff
@@ -112,11 +112,6 @@ final class TypeSyntaxVisitorSpec: QuickSpec {
         }
         """
 
-        it("detects the type name") {
```
bachand, line 73: The `TypeSyntaxAnalyzer` is initializer with a type name. As a result, we no longer return the type name in the analyzed result (i.e. we removed `TypeProperties`). And so we no longer need these tests.

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
```diff
@@ -222,24 +222,77 @@ final class TypeSyntaxVisitorSpec: QuickSpec {
...
+            public class FakeType {
+              var thing: String = "Hello, World"
+            }
+
+            extension FakeType {
+              var anotherThing: String = "Goodbye, World"
+            }
+            """
+
+            it("detects properties in the declaration and extension") {
```
bachand, line 261: It was great this just worked :)

Sources/SwiftInspectorAnalyzers/PropertyAnalyzer.swift
```diff
@@ -21,292 +21,18 @@
 // SOFTWARE.
```
bachand, line 21: Putting this here so we have a thread. Can I get access to CloudCov?

Screen Shot 2021-04-08 at 1 04 07 PM

fdiaz: You should have access https://app.codecov.io/gh/fdiaz/SwiftInspector

The links on the report don't work for me either 🤔

bachand: That did work, though this should def be easier. Thanks Dan.

bachand:
Btw if the reviewers prefer I can try to break this down into two to three PRs. The size definitely got a way from me a bit 😅 Just lmk.

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
```diff
@@ -267,7 +271,7 @@ extension TypeProperties.MergeError {
 }
 
 extension TypeProperties {
-  func merge(with other: TypeProperties?) throws -> TypeProperties {
+  public func merge(with other: TypeProperties?) throws -> TypeProperties {
```
dfed: Note to self while I review: this extension method is used by the analyzer, so it should also be `public`.

Sources/SwiftInspectorAnalyzers/Tests/PropertyAnalzyerSpec.swift
```diff
@@ -28,6 +28,7 @@ import Foundation
 import SwiftInspectorTestHelpers
 
 @testable import SwiftInspectorAnalyzers
+@testable import SwiftInspectorVisitors
```
dfed: curious if we really need a testable import here

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
dfed: at least that's how I liked to write the visitors. I approve of this method 🙂

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
```diff
@@ -247,7 +247,7 @@ public struct TypeProperties: Hashable {
   /// The name of the type.
   public let name: String
   /// The properties on this type
-  public let properties: [PropertyData]
+  public let properties: Set<PropertyData>
```
dfed: Makes sense that each property should be unique. Curious if/how you envision looking up the property data? I would think a dictionary might be more interesting so we could do a fast property name -> property data lookup.

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
```diff
@@ -31,68 +31,60 @@ final class TypeSyntaxVisitorSpec: QuickSpec {
...
-      }
+          let result = TypeSyntaxVisitor.merge(
+            [newPropertyData],
+            into: existingPropertiesData)
 
-      context("when the other type is nil") {
-        it("returns the original") {
-          expect(try? type1.merge(with: nil)) == type1
+          expect(result.count).to(equal(3))
+          expect(result.contains(newPropertyData)).to(beTrue())
```
dfed, line 84: I do see why the `Set` result type is useful in tests 🙂

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
dfed: I don't fully understand why `TypeSyntaxVisitor` is scoped to a single type name. I know it was doing that before, but... I don't fully get the use case. A `///` comment on `TypeSyntaxVisitor` might be helpful. I think it's finding the names of all declared types in a particular tree. Maybe what I'm looking for more than a `///` is a rename: `DeclaredTypeSyntaxVisitor`?

Open to thoughts here. None of these ideas are blocking since we're improving what was already here.

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
dfed: Tip to Francisco if he reviews after me: review this commit with whitespace changes ignored.

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
```diff
@@ -100,8 +100,8 @@ final class TypeSyntaxVisitorSpec: QuickSpec {
                       """
 
         it("returns type info with empty property list") {
-          let result = try? sut.analyze(fileURL: fileURL)
-          expect(result?.properties).to(beEmpty())
+          try VisitorExecutor.walkVisitor(sut, overContent: content)
```
dfed, line 103: well ain't that nice looking 🙂

dfed (approved):
Looks great! Separate commits made this super easy to review.

Sources/SwiftInspectorAnalyzers/Tests/PropertyAnalzyerSpec.swift
bachand: We eventually delete this file 👌

bachand: And in the new file we use the `@testable` to test the internal `merge(...)` method.

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
bachand: Currently (and before) we need to iterate the sequence to find an individual property.

A set seemed like a nice improvement since the code was [already](https://github.com/fdiaz/SwiftInspector/blob/2a901892629c14e828e165c58bc530d505494a30/Sources/SwiftInspectorAnalyzers/PropertyAnalyzer.swift) casting the properties to a set.

A dictionary is an interesting idea but a complication is that you can have a static and instance property of the same name, so the key can't be only the property name.

It's hard for me to know what a the most common workflow for this data would be: looking at all properties or finding a specific property.

Even though finding a specific property is O(N) I think it's reasonable to assume that N will be fairly small in this case.

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
bachand: Re: `DeclaredTypeSyntaxVisitor` it will find properties in an extension of a type too, [even if the extension is not on the type's declaration](https://github.com/fdiaz/SwiftInspector/blob/161075911041a9fbd24dc5f18dc042d10658be9d/Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift).

bachand: I also want to highlight [this test](https://github.com/fdiaz/SwiftInspector/blob/161075911041a9fbd24dc5f18dc042d10658be9d/Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift#L133-L166). `TypeSyntaxVisitor` will find all properties of a non fully qualified type name in a file, even if those properties are on multiple distinct types.

bachand: I think it would be really nice if we updated this visitor to accept a fully qualified type name. Until then, maybe we start with some documentation? 3ac70f3

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
fdiaz: Yeah, I've come back and forth on this honestly. I liked that the Visitors were fairly decoupled originally and was hoping we could use composition at some point to reuse them without having them know about the types (and having someone that calls into them know instead), but I really like the approach we took of using composition _while having these visitors know about the types they're generating_ I think it makes it really powerful.

dfed: Just gonna say I'm super happy with how easy https://github.com/fdiaz/SwiftInspector/pull/88 turned out to be. We're developing quite the toolset.

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
```diff
@@ -31,68 +31,60 @@ final class TypeSyntaxVisitorSpec: QuickSpec {
...
+            let result = TypeSyntaxVisitor.merge([newPropertyData], into: nil)
+            expect(result).to(equal([newPropertyData]))
+          }
+        }
+      }
+
+      context("when there is existing data about properties") {
+
+        it("merges the existing and new data") {
+          let newPropertyData = PropertyData(
```
fdiaz, line 61: nit: This setup is probably better places in a beforeEach block under the context?

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
```diff
@@ -31,68 +31,60 @@ final class TypeSyntaxVisitorSpec: QuickSpec {
...
+          let result = TypeSyntaxVisitor.merge(
+            [newPropertyData],
+            into: existingPropertiesData)
 
-      context("when the other type is nil") {
-        it("returns the original") {
-          expect(try? type1.merge(with: nil)) == type1
+          expect(result.count).to(equal(3))
+          expect(result.contains(newPropertyData)).to(beTrue())
+          expect(result.isSuperset(of: existingPropertiesData)).to(beTrue())
```
fdiaz, line 85: nit: 3 individual `it` instead of 3 expectations in one test?

Sources/SwiftInspectorAnalyzers/PropertyAnalyzer.swift
```diff
@@ -24,28 +24,15 @@ import Foundation
 import SwiftInspectorVisitors
 import SwiftSyntax
 
-// MARK: - PropertyAnalyzer
+extension StandardAnalyzer {
```
fdiaz, line 27: rename file?

fdiaz (approved):

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
bachand: I don't think we should _never_ write a "generalized" visitor. For example, regarding https://github.com/fdiaz/SwiftInspector/pull/87#discussion_r610324857, I think we eventually want to have a visitor that looks for a fully qualified type name. And then we'd have a more specific visitor inspect the properties of the types once they are found. But I don't think every visitor needs to be generalized.

Sources/SwiftInspectorAnalyzers/PropertyAnalyzer.swift
bachand: I followed [this](https://github.com/fdiaz/SwiftInspector/blob/24e21504da3fb5f55e1d041f8085a2e123f56254/Sources/SwiftInspectorAnalyzers/ImportsAnalyzer.swift) example.

Sources/SwiftInspectorVisitors/TypeSyntaxVisitor.swift
dfed: dig the doc comment

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
bachand: Since these are `let`s I didn't create a `beforeEach(...)` but I moved the values into the context. ad3e165

Sources/SwiftInspectorVisitors/Tests/TypeSyntaxVisitorSpec.swift
bachand: I simplified to one expectation. 7652f0e

PR #88, opened 2021-04-09
Find generic parameters on nestable types

dfed:
This PR adds the ability to find generic parameters defined on enums, structs, and classes.

### DIFF
--- a/Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
@@ -67,6 +67,22 @@ final class NestableTypeVisitorSpec: QuickSpec {
             }
           }
 
+          context("with a single generic") {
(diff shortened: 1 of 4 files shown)


Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
```diff
@@ -233,6 +265,22 @@ final class NestableTypeVisitorSpec: QuickSpec {
             }
           }
 
+          context("with a single generic") {
```
bachand, line 268: Do we have any tests for multiple generic parameters? Could be nice to have at least one.

bachand (approved):

Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
dfed: Good point! 04baaa1

PR #89, opened 2021-04-09
Encode rich Array and Dictionary information in TypeDescription

dfed:
When I originally wrote `TypeDescription`, I took a shortcut. Instead of creating `case array` and `case dictionary` types, I encoded these types in a canonical `simple` form.

This meant that `[Int]` would be stored as `.simple(name: "Array<Int>")` and `[Int: String]` would be stored as `.simple(name: "Dictionary<Int, String>")`.

This was fine at the time. Now that I'm trying to inspect the underlying generic type, I'm realizing we need to richly encode these types.

This PR introduces `case array` and `case dictionary`. I've included comments that highlight a different possible approach – curious to get the reviewers' thoughts.

### DIFF
--- a/Sources/SwiftInspectorVisitors/TypeDescription.swift
+++ b/Sources/SwiftInspectorVisitors/TypeDescription.swift
@@ -36,6 +36,10 @@ public enum TypeDescription: Codable, Hashable {
   indirect case optional(TypeDescription)
   /// An implicitly unwrapped optional type. eg. Int!
   indirect case implicitlyUnwrappedOptional(TypeDescription)
+  /// An array. e.g. [Int]
+  indirect case array(element: TypeDescription)
+  /// A dictionary. e.g. [Int: String]
+  indirect case dictionary(key: TypeDescription, value: TypeDescription)
   /// A tuple. e.g. (Int, String)
   indirect case tuple([TypeDescription])
   /// A type that can't be represented by the above cases.
@@ -99,6 +103,10 @@ public enum TypeDescription: Codable, Hashable {
       } else {
         return "\(parentType.asSource).\(name)<\(generics.map { $0.asSource }.joined(separator: ", "))>"
       }
+    case let .array(element):
+      return "Array<\(element.asSource)>"
+    case let .dictionary(key, value):
+      return "Dictionary<\(key.asSource), \(value.asSource)>"
     case let .tuple(types):
       return "(\(types.map { $0.asSource }.joined(separator: ", ")))"
     case let .unknown(text):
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -247,18 +278,12 @@ extension TypeSyntax {
       return .implicitlyUnwrappedOptional(typeIdentifier.wrappedType.typeDescription)
 
     } else if let typeIdentifier = self.as(ArrayTypeSyntax.self) {
-      // An array of the form [Type] is synonymous with Array<Type>.
-      // Utilize a simple type rather than creating another enum case.
-      let elementDescription = typeIdentifier.elementType.typeDescription.asSource
-      return .simple(name: "Array<\(elementDescription)>")
+      return .array(element: typeIdentifier.elementType.typeDescription)
```
dfed, line 281: Another possible option here is to encode this type as: 
```suggestion
      return .simple(name: "Array", generics: [typeIdentifier.elementType.typeDescription])
```

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -247,18 +278,12 @@ extension TypeSyntax {
...
     } else if let typeIdentifier = self.as(DictionaryTypeSyntax.self) {
-      // An array of the form [KeyType: ValueType] is synonymous with Dictionary<KeyType, ValueType>.
-      // Utilize a simple type rather than creating another enum case.
-      let keyDescription = typeIdentifier.keyType.typeDescription
-      let valueDescription = typeIdentifier.valueType.typeDescription
-      return .simple(
-        name: "Dictionary<\(keyDescription.asSource), \(valueDescription.asSource)>")
+      return .dictionary(
+        key: typeIdentifier.keyType.typeDescription,
+        value: typeIdentifier.valueType.typeDescription)
```
dfed, line 286: Another possible option here is to encode this type as: 
```suggestion
      return .simple(
        name: "Dictionary",
        generics: [typeIdentifier.keyType.typeDescription, typeIdentifier.valueType.typeDescription])
```
This would allow us to get rid of the `.array` case. We'd lose the context that we _know_ that we're dealing with a `Foundation.Array` type, but this way we'd ensure that `Array<Int>` and `[Int]` are treated the same way by the system. Thoughts?

Sources/SwiftInspectorVisitors/TypeDescription.swift
fdiaz: I actually prefer the .array case tbh.

fdiaz (approved):

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I guess we can always have consuming code make this mapping if we really want it. 👍

bachand: I know I was involved in discussions about this previously and I can't recall exactly where I stood. Moving forward I do agree we should have the `.array` and `.dictionary` cases and I also do think having a way to normalize at a higher level could be cool.

bachand (commented):
Looked over at a high level but deferring to Francisco's review.

PR #90, opened 2021-04-13
Enable nestable type visitor to find generic requirements

dfed:
Nestable type declarations can have generic `where` requirements. This PR enables us to find those requirements.

Best reviewed commit-by-commit.

### DIFF
--- a/Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
@@ -201,7 +201,7 @@ final class NestableTypeVisitorSpec: QuickSpec {
               expect(structInfo?.genericParameters.map { $0.name }) == ["A", "B"]
             }
 
-            context("that is constrained with a where clause") {
+            context("where one parameter is constrained by a where clause") {
               it("finds the constraint") {
                 let content = """
                 public struct SomeStruct<A, B>: SomeProtocol where A: CustomStringConvertible {
@@ -320,7 +320,7 @@ final class NestableTypeVisitorSpec: QuickSpec {
               expect(enumsInfo?.genericParameters.map { $0.name }) == ["A", "B", "C"]
             }
 
-            context("that is constrained with a where clause") {
+            context("where one parameter is constrained by a where clause") {
               it("finds the constraint") {
                 let content = """
                 public enum SomeEnum<A, B, C>: SomeProtocol where A: CustomStringConvertible {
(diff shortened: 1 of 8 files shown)


Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
```diff
@@ -180,6 +200,26 @@ final class NestableTypeVisitorSpec: QuickSpec {
               expect(structInfo?.name) == "SomeStruct"
               expect(structInfo?.genericParameters.map { $0.name }) == ["A", "B"]
             }
+
+            context("that is constrained with a where clause") {
```
bachand, line 204: "with one that is..."?

bachand: (since the parent context is "with two generic parameters")

bachand: Same comment below.

bachand (approved):

Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
```diff
@@ -81,6 +81,26 @@ final class NestableTypeVisitorSpec: QuickSpec {
               expect(classInfo?.name) == "SomeClass"
               expect(classInfo?.genericParameters.map { $0.name }) == ["T"]
             }
+
+            context("that is constrained with a where clause") {
```
bachand, line 85: Would it be worth having a test somewhere to verify that we can find generic constraints when the type with the generic constraints is nested? Not sure it's valuable, so legitimately curious.

```swift
enum Parent {
  public class SomeClass<T>: SomeProtocol where T: CustomStringConvertible {
    public typealias SomeProtocolAssociatedType = A
  }
}
```

dfed: I don't think we need this since the generic requirement is a property on the nestable type. We have tests showing that:
1. Deeply nested types are found and given the correct names
2. We can parse/find/surface generic requirements on nestable types

Between those two tests and the fact that our nested types are found via recursion, we can use induction to know that we can parse/find/surface generic requirements on deeply nested types.

Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
dfed: 87bf5a3

fdiaz (commented):
Good stuff!

PR #91, opened 2021-04-19
FileVisitor should surface inner typealiases on protocol definitions

dfed:
Just missing a little bit of glue code


bachand (commented):
Nice!

PR #92, opened 2021-04-19
Expose Test Helpers as a library and stop exposing Commands

fdiaz:
This PR just changes what we expose on SwiftInspector externally. We don't need to expose the Commands, but exposing the Test Library can be useful since there's some niceness there for making temp files.


dfed (approved):
I don't love that changes to SwiftInspectorTestHelpers could force a major version bump. But I agree that what we've written there is helpful so I can swallow that pill.

PR #97, opened 2021-07-07
Improve VisitorExecutor

fdiaz:
**Easier to review commit-by-commit**

A minor refactor on VisitorExecutor to make it easier to read. My thinking here is that I prefer to have the tests act on the subject under test rather than passing the subject under test to a method which feels less natural to me.

I also got rid of all the temporary file creation / deletion since we can use `let syntax: SourceFileSyntax = try SyntaxParser.parse(source: content)` instead.


Sources/SwiftInspectorTestHelpers/SyntaxVisitor+WalkContent.swift
```diff
@@ -25,21 +25,13 @@
...
-    _ visitor: Visitor,
-    overContent content: String)
-  throws
-  {
-    let fileURL = try Temporary.makeFile(content: content)
-    let syntax: SourceFileSyntax = try SyntaxParser.parse(fileURL)
-    visitor.walk(syntax)
-    try Temporary.removeItem(at: fileURL)
+  public func walkContent(_ content: String) throws {
+    let syntax: SourceFileSyntax = try SyntaxParser.parse(source: content)
```
dfed, line 34: I don't love that we can't write a test for this new method but... I guess everything will fail if this fails

Sources/SwiftInspectorTestHelpers/SyntaxVisitor+WalkContent.swift
```diff
@@ -25,21 +25,13 @@
...
+    self.walk(syntax)
```
dfed, line 35: do we need `self.` here?

dfed (approved):

Sources/SwiftInspectorTestHelpers/SyntaxVisitor+WalkContent.swift
fdiaz: We do not!

Sources/SwiftInspectorTestHelpers/SyntaxVisitor+WalkContent.swift
fdiaz: I'm fine not writing a test for this since that'd test that `SyntaxParser` is working appropriately, which we shouldn't really test IMO.

PR #99, opened 2021-07-08
Properties: Handle edge case

fdiaz:
We were not handling the 

```swift
let red, green, blue: Double
let seconds: Int, hours, years: Double
```

cases correctly.

This PR adds unit tests + handles this use case.

### DIFF
--- a/Sources/SwiftInspectorVisitors/PropertySyntaxVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/PropertySyntaxVisitor.swift
@@ -33,30 +33,20 @@ public final class PropertySyntaxVisitor: SyntaxVisitor {
   public override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
     let modifier = findModifiers(from: node)
 
-    node.bindings.forEach { binding in
+    var lastFoundType: String?
+    node.bindings.reversed().forEach { binding in
       if let identifier = binding.pattern.as(IdentifierPatternSyntax.self) {
-        if
-          let typeAnnotation = binding.typeAnnotation,
-          let simpleTypeIdentifier = typeAnnotation.type.as(SimpleTypeIdentifierSyntax.self)
-        {
-          propertiesInfo.insert(.init(
-                                  name: identifier.identifier.text,
-                                  typeAnnotation: simpleTypeIdentifier.name.text,
-                                  modifiers: modifier))
-          return
-        }
-        // no type annotation was found. In this case it's much more complicated to get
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorVisitors/PropertySyntaxVisitor.swift
```diff
@@ -102,6 +92,19 @@ public final class PropertySyntaxVisitor: SyntaxVisitor {
 
     return modifier
   }
+
+  private func findTypeName(from node: TypeAnnotationSyntax?) -> String? {
+    guard
```
fdiaz, line 97: This is basically a copy/paste from https://github.com/fdiaz/SwiftInspector/pull/99/files#diff-8622b0ddc6698d01f425675d628fce88a76b661f30c567a9756e1282c9981e87L38

Sources/SwiftInspectorVisitors/PropertySyntaxVisitor.swift
```diff
@@ -33,30 +33,20 @@ public final class PropertySyntaxVisitor: SyntaxVisitor {
   public override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
     let modifier = findModifiers(from: node)
 
-    node.bindings.forEach { binding in
+    var lastFoundType: String?
+    node.bindings.reversed().forEach { binding in
```
fdiaz, line 37: We're starting from the bottom ~now we're here~ to find the last type found.

Sources/SwiftInspectorVisitors/PropertySyntaxVisitor.swift
```diff
@@ -33,30 +33,20 @@ public final class PropertySyntaxVisitor: SyntaxVisitor {
...
-        // TODO: Include logic to get types of any variable declaration
+        let typeName = findTypeName(from: binding.typeAnnotation)
+        // If we find a type name on the n element, but n-1 doesn't have an explicit type name, n-1 has the same type as n
+        // e.g. let red: Int, green, blue: Double
+        // where both green and blue are of type Double
+        if let typeName = typeName { lastFoundType = typeName }
         propertiesInfo.insert(.init(
                                 name: identifier.identifier.text,
-                                typeAnnotation: nil,
+                                typeAnnotation: typeName ?? lastFoundType,
```
dfed, line 46: since we do `if let typeName = typeName { lastFoundType = typeName }` on the line above, can't we just do:
```suggestion
                                typeAnnotation: lastFoundType,
```

dfed (approved):

PR #100, opened 2021-07-09
Use properties in Protocols / Extensions and Nested Types

fdiaz:
**Easier to review commit-by-commit**

This PR includes information about properties in both protocols / extensions and nested types.

PR #102, opened 2021-07-09
Do not propagate properties to parents of nested types

dfed:
We introduced a bug in #100 that I caught after the PR landed where a parent of a nested type that vended a property would have that property assigned to both the parent and nested type.

This PR fixes that issue. **Best reviewed in Files changed**.

### DIFF
--- a/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
@@ -62,8 +62,6 @@ public final class PropertyVisitor: SyntaxVisitor {
     if visitTypeDeclarationChildren {
       return .visitChildren
     } else {
-      // We've encountered a nested type declaration.
-      // Skip it to avoid scanning a child type's properties.
       return .skipChildren
     }
   }
@@ -72,8 +70,6 @@ public final class PropertyVisitor: SyntaxVisitor {
     if visitTypeDeclarationChildren {
       return .visitChildren
     } else {
-      // We've encountered a nested type declaration.
-      // Skip it to avoid scanning a child type's properties.
       return .skipChildren
     }
   }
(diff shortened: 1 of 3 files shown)


Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -58,6 +58,36 @@ public final class PropertyVisitor: SyntaxVisitor {
...
+
+  public override func visit(_ node: EnumDeclSyntax) -> SyntaxVisitorContinueKind {
+    if visitTypeDeclarationChildren {
+      return .visitChildren
+    } else {
+      // We've encountered a nested type declaration.
+      // Skip it to avoid scanning a child type's properties.
+      return .skipChildren
+    }
+  }
```
dfed, line 89: Thoughts on whether I should also stop on `Protocol` and `Extension` syntax nodes? Those aren't nestable, but they are type declarations... aaaaand I just talked myself into it more commits coming.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -51,6 +51,26 @@ public final class PropertyVisitor: SyntaxVisitor {
     return .skipChildren
   }
 
+  public override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
```
fdiaz, line 54: Should we also assert here to keep it consistent with other types?

fdiaz (approved):
Nice!

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
dfed: We don't need to assert – it's valid for this visitor to encounter decls. I only assert when we encounter a decl that wouldn't compile, like seeing any type decl within a protocol's `members` node.

I'm not convinced that asserting is actually a useful thing to do there since it's reasonable to run `SwiftInspector` over code that doesn't compile. But we can tackle that later

PR #103, opened 2021-07-09
Traverse node.members only once

### LINKED ISSUE #101: Create single MemberVisitor type that walks MemberDeclBlockSyntax nodes
Right now we have `TypealiasVisitor`, `PropertyVisitor`, and `AssociatedtypeVisitor` that each walk a `MemberDeclBlockSyntax` node.

Ideally we wouldn't walk a single `MemberDeclBlockSyntax` node multiple times, and instead have a single `MemberVisitor` that walks these node types and then defers to a `TypealiasVisitor`, `PropertyVisitor`, and `AssociatedtypeVisitor` whenever the nodes these types care about are hit.

This proposed structure will likely yield a small performance improvement.

dfed:
Resolves #101

### DIFF
--- a/Sources/SwiftInspectorVisitors/ProtocolVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/ProtocolVisitor.swift
@@ -94,7 +94,6 @@ public final class ProtocolVisitor: SyntaxVisitor {
 
     // We don't need to visit children because our visitor just did that for us.
     return .skipChildren
-
   }
 
   public override func visit(_ node: TypealiasDeclSyntax) -> SyntaxVisitorContinueKind {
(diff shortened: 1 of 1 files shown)


Sources/SwiftInspectorVisitors/ProtocolVisitor.swift
```diff
@@ -90,9 +87,45 @@ public final class ProtocolVisitor: SyntaxVisitor {
...
   }
 
+  public override func visit(_ node: AssociatedtypeDeclSyntax) -> SyntaxVisitorContinueKind {
+    let associatedtypeVisitor = AssociatedtypeVisitor()
+    associatedtypeVisitor.walk(node)
+    associatedtypes.append(contentsOf: associatedtypeVisitor.associatedTypes)
+
+    // We don't need to visit children because our visitor just did that for us.
+    return .skipChildren
+
```
dfed, line 97:
```suggestion
```

fdiaz (approved):

PR #104, opened 2021-07-27
[iOS] Use TypeDescription in PropertyInfo

fdiaz:
**Easier to review commit-by-commit**

This PR moves us into using TypeDescription for PropertyInfo instead of a simple String so we can get more information about the type.

I noticed that we were not supporting `open` properties either, so I went ahead and added those too.


dfed (approved):
🎉

PR #106, opened 2021-09-11
Do not use 'expect(...).to(throwAssertion())'

dfed:
This test method [crashes on M1 Macs](https://github.com/Quick/Nimble/blob/cfa125829821415bfaa9a43f03ee1cac6f743715/Sources/Nimble/Matchers/ThrowAssertion.swift#L124-L128), so instead of testing that we throw an assertion, this PR creates a way for us to test that we intend to assert.

Instead of calling `Swift.assertionFailure`, we'll now call `SwiftInspectorVisitors.assertionFailureOrPostNotification`, which will either assert or post a notification based on a static boolean.

The use of an static boolean with no threading barrier in this project is okay because the entire repo is single-threaded.

**We will need to merge this PR through a failing CodeCov check** because [we are expectedly not executing the `assertionFailure` line](https://app.codecov.io/gh/fdiaz/SwiftInspector/compare/106/diff#D3L33) in tests.

### DIFF
--- a/Sources/SwiftInspectorVisitors/AssertionFailure.swift
+++ b/Sources/SwiftInspectorVisitors/AssertionFailure.swift
@@ -1,0 +1,43 @@
+// Created by Dan Federman on 9/11/21.
+// Copyright © 2021 Dan Federman
+//
+// Distributed under the MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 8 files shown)


Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -294,7 +294,7 @@ extension TypeSyntax {
       return .simple(name: "AnyObject")
 
     } else {
-      assertionFailure("TypeSyntax of unexpected type. Defaulting to `description`.")
+      assertionFailureOrPostNotification("TypeSyntax of unexpected type. Defaulting to `description`.")
```
dfed, line 297: We previously had no tests for this line.

Sources/SwiftInspectorVisitors/AssertionFailure.swift
```diff
@@ -0,0 +1,43 @@
...
+import Foundation
+
+/// Indicates that an internal confidence check failed.
+/// Posts the `AssertionFailure.notification` if `AssertionFailure.postNotification` is `true`, otherwise calls `Swift.assertionFailure`.
+/// Using this method instead of `Swift.assertionFailure` allows for testing that assertions are triggered.
+@inlinable public func assertionFailureOrPostNotification(_ message: @autoclosure () -> String = String(), file: StaticString = #file, line: UInt = #line) {
+  if AssertionFailure.postNotification {
+    NotificationCenter.default.post(AssertionFailure.notification)
+  } else {
+    assertionFailure(message(), file: file, line: line)
```
dfed, line 33: It is expected that we don't execute this line during testing.

bachand (approved):
Have we checked if a future version of Nimble fixes this? I wonder if they will need to remove the `throwAssertion()` API for M1 Macs?

dfed:
> Have we checked if a future version of Nimble fixes this?

I did and it doesn't.

> I wonder if they will need to remove the throwAssertion() API for M1 Macs?

It'd be lovely for this failure to be compile-time rather than run-time. I haven't filed an issue though.

PR #107, opened 2021-09-11
Make TypeDescription support FunctionTypeSyntax

### LINKED ISSUE #105: Make TypeDescription handle FunctionTypeSyntax
Right now running `FileVisitor` over any file that has a property that has a closure type causes an assertion. We should fix this.

dfed:
Resolves #105

### DIFF
--- a/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
@@ -184,6 +184,41 @@ final class TypeDescriptionSpec: QuickSpec {
     }
     """.data(using: .utf8)!
 
+  let closureTestCase = TypeDescription.closure(
+    arguments: [
+      .simple(name: "Foo"),
+      .optional(.simple(name: "Bar"))
+    ],
+    doesThrow: true,
+    returnType: .simple(name: "FooBar"))
+  let closureTestCaseData = """
+    {
+      "caseDescription": "closure",
+      "typeDescriptions": [
+        {
+          "caseDescription": "simple",
+          "text": "Foo",
+          "typeDescriptions": []
+        },
+        {
+          "caseDescription": "optional",
+          "typeDescription":
+          {
+            "caseDescription": "simple",
+            "text": "Bar",
+            "typeDescriptions": []
+          }
+        }
+      ],
+      "text": "throws",
+      "typeDescription": {
+        "caseDescription": "simple",
+        "text": "FooBar",
+        "typeDescriptions": []
+      }
+    }
+    """.data(using: .utf8)!
+
   let unknownTestCase = TypeDescription.unknown(text: "Foo")
   let unknownTestCaseData = """
     {
@@ -243,6 +278,12 @@ final class TypeDescriptionSpec: QuickSpec {
       context("that represents a tuple type") {
         it("decodes the encoded type description") {
           expect(try decoder.decode(TypeDescription.self, from: self.tupleTestCaseData)) == self.tupleTestCase
+        }
+      }
+
+      context("that represents a closure type") {
+        it("decodes the encoded type description") {
+          expect(try decoder.decode(TypeDescription.self, from: self.closureTestCaseData)) == self.closureTestCase
         }
       }
 
@@ -316,6 +357,12 @@ final class TypeDescriptionSpec: QuickSpec {
       context("utilizing a tuple type") {
         it("successfully decodes the data") {
           expect(try decoder.decode(TypeDescription.self, from: try encoder.encode(self.tupleTestCase))) == self.tupleTestCase
+        }
+      }
+
+      context("utilizing a closure type") {
+        it("successfully decodes the data") {
+          expect(try decoder.decode(TypeDescription.self, from: try encoder.encode(self.closureTestCase))) == self.closureTestCase
         }
       }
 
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -158,6 +161,12 @@ public enum TypeDescription: Codable, Hashable {
       let typeDescriptions = try values.decode([Self].self, forKey: .typeDescriptions)
       self = .tuple(typeDescriptions)
 
+    case Self.closureDescription:
+      let typeDescriptions = try values.decode([Self].self, forKey: .typeDescriptions)
+      let text = try values.decodeIfPresent(String.self, forKey: .text)
+      let typeDescription = try values.decode(Self.self, forKey: .typeDescription)
```
dfed, line 167: Thoughts on re-using these existing keys? It's a little hacky, but it does work.

Biggest benefit of introducing new keys is to create a key that records a `Bool` rather than `String`.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -186,11 +195,17 @@ public enum TypeDescription: Codable, Hashable {
...
+      if doesThrow {
+        try container.encode("throws", forKey: .text)
+      }
+      try container.encode(returnType, forKey: .typeDescription)
     }
   }
 
   enum CodingKeys: String, CodingKey {
-    /// The value for this key is the case encoded as a string.
+    /// The value for this key is the case encoded as a String.
```
dfed, line 208: This line change is unrelated to the rest of the PR but figured I'd make the capitalization consistent while I was here.

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -684,6 +731,51 @@ final class TypeDescriptionSpec: QuickSpec {
...
+            functionIdentifier = node.type.as(TypeSyntax.self)?.typeDescription
+            return .skipChildren
+          }
+        }
+
+        context("on a function that does not throw") {
+          var visitor: FunctionTypeSyntaxVisitor!
+          beforeEach {
+            let content = """
+                var test: (Int, Double) -> String
```
bachand, line 750: Do we need to handle `@escaping`?

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: I had a question above about `@escaping`. How we handle that could impact how we approach new keys.

A key for boolean doesn't sit right with me: we'd be creating a general-purpose key for a single existing use case. It's not clear to me that we'd use the key again.

If we are going to create a new key I'd recommend we create one that is specific to this use case. I don't think there is much downside to more semantically meaningful keys.

I also wonder if we could use the `caseDescription` to decode sub-objects that are strongly typed.

```swift
enum CodingKeys: String, CodingKey {
  case caseDescription
  case typeData
}
```

Then we would define a simple `Codable` struct for each object with the properties we want, and store the encoded data for `typeData` key.

So for the closure case we could imagine defining a struct like this.

```swift
struct Closure: Codable {
  let parameterTypes: [TypeDescription]
  let returnType: [TypeDescription]
  let isEscaping: Bool
  let doesThrow: Bool
}
```

`TypeDescription` is becoming complex enough that the additional structure could be valuable. We would need to understand if there is a major increase in decoding/encoding time.

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
bachand: Down the road we will also need to handle `async`.

bachand (approved):

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
dfed: We will need to handle `async`, but `FunctionTypeSyntax` [doesn't actually have a way to access `@escaping`](https://github.com/apple/swift-syntax/blob/ad768ea60c913afc15bab05e3a3d2ef77a46beba/Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxTypeNodes.swift#L1442). I'm not sure what syntax type captures that information.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: > A key for boolean doesn't sit right with me: we'd be creating a general-purpose key for a single existing use case. It's not clear to me that we'd use the key again.

Entirely fair. I agree with you the creating specific keys for our use case would be a reasonable approach.

I like the idea of creating `Codable` structs for each enum case and using those to serialize/deserialize, but that feels out of scope for this PR. @bachand can you create an issue for this? Seems like a good improvement that we could tackle in one fell swoop. I don't expect using strongly-typed `Codable` structs would negatively impact performance.

Only downside I can think of to that change is that it'd technically be a breaking change. But hey, the repo issue still in beta so nbd.

dfed: Used new keys in 728b500

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -42,6 +42,8 @@ public enum TypeDescription: Codable, Hashable {
   indirect case dictionary(key: TypeDescription, value: TypeDescription)
   /// A tuple. e.g. (Int, String)
   indirect case tuple([TypeDescription])
+  /// A closure. e.g. (Int, Double) throws -> String
+  indirect case closure(arguments: [TypeDescription], doesThrow: Bool, returnType: TypeDescription)
```
fdiaz, line 46: Instead of `doesThrow` should this be something like `keywords` so we can extend it later on? It could be something like an OptionSet, and then we could have a convenience parameter for `doesThrow`.

fdiaz (approved):

Sources/SwiftInspectorVisitors/TypeDescription.swift
fdiaz: I fully agree with @bachand about the usage of a struct here and moving away from the enum.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I think as we add more types we can talk about simplifying, but we've got only one potential new parameter in the future that we know of so I'm hesitant to optimize just yet.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -217,6 +215,12 @@ public enum TypeDescription: Codable, Hashable {
     case dictionaryKey
     /// The value for this key is a dictionary's value of type TypeDescription
     case dictionaryValue
+    /// The value for this key represents the list of types in a closure argument list and is of type [TypeDescription]
```
bachand, line 218: 👍

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: https://github.com/fdiaz/SwiftInspector/issues/109

PR #110, opened 2021-09-15
Record let/var and trailing source code for properties

bachand:
Best reviewed commit by commit.

`PropertyInfo` include a new `Paradigm` which will tell you whether the property is a `let` or `var`. Furthermore it will inform you if the property is undefined at its declaration, initialized at its declaration, or computed. When the property is initialized at its declaration or computed we include the source-accurate text of the initializer or computation code block, respectively.

I implemented a `Codable` conformance to `PropertyInfo.Paradigm` as well.

@dfed cc: @fdiaz


Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -74,23 +76,24 @@ public final class PropertyVisitor: SyntaxVisitor {
   // MARK: Private
 
   private func findModifiers(from node: VariableDeclSyntax) -> PropertyInfo.Modifier {
```
bachand, line 78: A couple of interesting findings that I had.
Screen Shot 2021-09-15 at 1 11 25 PM
Screen Shot 2021-09-15 at 1 11 42 PM

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -69,10 +69,10 @@ final class PropertyVisitorSpec: QuickSpec {
               name: "red",
               typeDescription: .simple(name: "Double"),
               modifiers: [.internal, .instance]),
-              .init(
-                name: "blue",
-                typeDescription: .simple(name: "Double"),
-                modifiers: [.internal, .instance])
+            .init(
```
bachand, line 76: Used Xcode to indent.

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -274,13 +274,7 @@ final class PropertyVisitorSpec: QuickSpec {
...
         it("does not detect properties within type declaration") {
           try sut.walkContent(content)
-          let notExpected: [PropertyInfo] = [
-            .init(
-              name: "timestamp",
-              typeDescription: .simple(name: "Int"),
-              modifiers: [.internal, .instance])
-          ]
-          expect(sut.properties).notTo(contain(notExpected))
+          expect(sut.properties.map(\.name)).notTo(contain("timestamp"))
```
bachand, line 380: This test will be more robust if we only check the name.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -168,12 +168,30 @@ public struct PropertyInfo: Codable, Hashable, CustomDebugStringConvertible {
     }
   }
 
+  public enum Paradigm {
```
bachand: Dan—Francisco and I spoke further and aligned on this approach rather than including the nodes themselves in `PropertyInfo`. This approach allows `PropertyInfo` to remain a `Codable` struct. Moreover, we are able to capture some of the code information into this enum.

Sources/SwiftInspectorVisitors/PropertyInfo.swift
```diff
@@ -1,4 +1,90 @@
 // Created by Michael Bachand on 9/15/21.
 // Copyright © 2021 Airbnb Inc. All rights reserved.
 
-import Foundation
+// MARK: - PropertyInfo
+
+public struct PropertyInfo: Codable, Hashable, CustomDebugStringConvertible {
```
bachand, line 6: This type is big enough for its own file, especially considering that we are adding a custom `Codable` conformance that we will test.

Sources/SwiftInspectorVisitors/Tests/PropertyInfoSpec.swift
```diff
@@ -2,3 +2,131 @@
...
+import SwiftSyntax
+
+@testable import SwiftInspectorVisitors
+
+final class PropertyInfoParadigmSpec: QuickSpec {
+
+  let undefinedConstantTestCase = PropertyInfo.Paradigm.undefinedConstant
+  let undefinedConstantTestCaseData = """
+    {
+      "caseValue": 0
```
bachand, line 17: By encoding the cases as numbers rather than strings we make the encoded data smaller and reduce computation.

Sources/SwiftInspectorVisitors/Tests/PropertyInfoSpec.swift
```diff
@@ -2,3 +2,131 @@
...
 
 import Foundation
+import Nimble
+import Quick
+import SwiftInspectorTestHelpers
+import SwiftSyntax
+
+@testable import SwiftInspectorVisitors
+
+final class PropertyInfoParadigmSpec: QuickSpec {
```
bachand, line 12: I used TypeDescriptionSpec as a guide for these tests.

Sources/SwiftInspectorVisitors/PropertyInfo.swift
```diff
@@ -77,14 +77,84 @@ extension PropertyInfo {
     /// - Parameter codeBlockDesciption: A source-accurate description of the code block which computes the value.
     case computedVariable(_ codeBlockDesciption: String)
 
+    // MARK: Lifecycle
+
     public init(from decoder: Decoder) throws {
-      // TODO implement
-      fatalError()
+      let values = try decoder.container(keyedBy: CodingKeys.self)
```
bachand, line 83: I used the `TypeDescription` `Codable` implementation as a guide.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -76,23 +76,24 @@ public final class PropertyVisitor: SyntaxVisitor {
...
+            let leftParen = modifierSyntax.detailLeftParen,
+            let detail = modifierSyntax.detail,
+            let rightParen = modifierSyntax.detailRightParen
+          {
+            return modifierSyntax.name.text + leftParen.text + detail.text + rightParen.text
           }
-        result.append(contentsOf: modifiers)
-      }
+          return modifierSyntax.name.text
+        }
```
bachand, line 97: This code is copied.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -76,23 +76,24 @@ public final class PropertyVisitor: SyntaxVisitor {
...
-            if
-              let leftParen = modifierSyntax.detailLeftParen,
-              let detail = modifierSyntax.detail,
-              let rightParen = modifierSyntax.detailRightParen
-            {
-              return modifierSyntax.name.text + leftParen.text + detail.text + rightParen.text
-            }
-            return modifierSyntax.name.text
+    let modifiersString: [String]
+    if let modifiersSyntaxList = node.modifiers {
```
bachand, line 85: I simplified by using the `modifiers` accessor.

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -117,7 +117,7 @@ final class PropertyVisitorSpec: QuickSpec {
               name: "thing",
               typeDescription: .simple(name: "String"),
               modifiers: [.public, .static],
-              paradigm: .definedVariable("\"Hello, World\""))
+              paradigm: .definedVariable("= \"Hello, World\""))
```
bachand, line 120: I originally planned to remove this though when I wrote the code it felt more logical to include it since it is a part of the initializer syntax node.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -129,4 +134,15 @@ public final class PropertyVisitor: SyntaxVisitor {
     }
     return typeSyntax.typeDescription
   }
+
+  private func findPropertyType(from node: VariableDeclSyntax) -> PropertyType {
+    if let type = PropertyType(rawValue: node.letOrVarKeyword.text) {
+      return type
+    }
+    else {
+      assertionFailure("Property must be either let or var. The Swift language has evolved if you hit this assertion.")
```
bachand, line 177: It is expected that we do not test this assertion.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -135,6 +135,40 @@ public final class PropertyVisitor: SyntaxVisitor {
...
+      else {
+        return .undefinedVariable
+      }
+    }
+  }
+
+  private func findInitializerDescription(from node: PatternBindingListSyntax) -> String? {
+    let initializerClauseSyntaxes = node.compactMap { $0.initializer }
+    assert(initializerClauseSyntaxes.count <= 1, "A property should have at most one initializer.")
+    return initializerClauseSyntaxes.first?.description
```
bachand, line 163: `description` seems like the way to get back source code.

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -324,6 +324,23 @@ final class PropertyVisitorSpec: QuickSpec {
               paradigm: .definedVariable("= Foo()"))
           ]
         }
+
+        context("initialized with a closure") {
+          let content = """
+          var foo = { Foo() }()
```
bachand, line 330: It seemed like this could be an edge case.

bachand (commented):
We may need to merge through code coverage since we can't test an assertion failure that should never be executed.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
dfed: oooh I like this!

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -258,6 +258,91 @@ final class PropertyVisitorSpec: QuickSpec {
...
+          try sut.walkContent(content)
+          expect(sut.properties) == [
+            PropertyInfo(
+              name: "foo",
+              typeDescription: nil,
+              modifiers: [.internal, .instance],
+              paradigm: .computedVariable("{ Foo() }"))
+          ]
+        }
+      }
```
dfed, line 361: absolutely love these tests. Super simple to follow

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
dfed: I don't love including this constant prefix – I'm having trouble seeing how it's valuable.

I don't think it's problematic, but feels like `initializerDescription` already tells us that we're looking at the contents of an initializer so we don't need to repeat the `= `.

Maybe I'm missing something though. What led you to add this data?

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -135,6 +135,40 @@ public final class PropertyVisitor: SyntaxVisitor {
...
+      }
+      else {
+        return .undefinedVariable
+      }
+    }
+  }
+
+  private func findInitializerDescription(from node: PatternBindingListSyntax) -> String? {
+    let initializerClauseSyntaxes = node.compactMap { $0.initializer }
+    assert(initializerClauseSyntaxes.count <= 1, "A property should have at most one initializer.")
```
dfed, line 162: not sure if we care, but this will likely assert if you write code that won't compile. Example:
```
let foo = 5 = 3
```

dfed (approved):
Excited for this! I really like what y'all did with the paradigm. I like the `= ` storage less, but I can live with it.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
bachand: I will add a test and we can check.

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
bachand: When I was writing [this](https://github.com/fdiaz/SwiftInspector/blob/57ddcd73a5e7d684fd310c23bcf11012a24e4ca9/Sources/SwiftInspectorVisitors/PropertyVisitor.swift#L160-L164) method I found that the code to remove the syntax node with the `=` was a bit messy. Basically what I had was something like this:

```swift
return initializerClauseSyntaxes.first?.children?.last?.description
```

I had found that the children of `InitializerClauseSyntax` _seemed_ to be two nodes, the first of which contained the `=`. I was then starting to assert that `InitializerClauseSyntax` only contained two nodes, and it began to feel like brittle code.

I began to feel like I was making assumptions that may not be valuable for everyone using this library, and it felt like there was simplicity in `initializerDescription` containing the literal source of the `InitializerClauseSyntax` object.

As I thought further, one nice thing about the way we have it now is that a consumer of this visitor could theoretically reparse `initializerDescription` with SwiftSyntax at a higher layer and get back an `InitializerClauseSyntax`. That felt like a nice contract.

I could be convinced to remove the `=` (which we will do in our higher layer anyway) though I had trouble convincing myself that the added code here was adding a lot of value compared to the complexity, though I agree the tests are a bit annoying.

dfed: Got it! Okay I didn't realize that removing the `= ` added so much complexity and potential for failure points – I'm now entirely onboard with this approach.

bachand: Unfortunately reparsing the description won't give back an `InitializerClauseSyntax`. That being said I still think we stay simple until the range of use cases become more apparent.
image

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
bachand: Bizarre, but no assertion. https://github.com/fdiaz/SwiftInspector/pull/110/commits/1fbea486dc08f259ea7dcc8dbf1f72e1f446775d

Sources/SwiftInspectorVisitors/PropertyInfo.swift
```diff
@@ -0,0 +1,160 @@
...
+      case "open": self = .open
+      case "public": self = .public
+      case "private": self = .private
+      case "fileprivate": self = .fileprivate
+      case "private(set)": self = .privateSet
+      case "internal(set)": self = .internalSet
+      case "public(set)": self = .publicSet
+      case "internal": self = .internal
+      case "static": self = .static
+      default: self = []
```
bachand, line 56: This line was not covered before my PR, though I moved it.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
dfed: Interesting 🤔

bachand:
Merging through code coverage check due to assertion that will never trigger in normal circumstances.

Sources/SwiftInspectorVisitors/Tests/PropertyInfoSpec.swift
```diff
@@ -2,3 +2,131 @@
...
+      "caseValue": 0
+    }
+    """.data(using: .utf8)!
+
+  let definedConstantTestCase = PropertyInfo.Paradigm.definedConstant("Foo()")
+  let definedConstantTestCaseData = """
+    {
+      "caseValue": 1,
+      "initializerDescription": "Foo()"
+    }
```
fdiaz, line 26: I'd recommend creating a helper function that gives you this, that way if we ever need to update this you only do it at one place cc @bachand

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -258,6 +258,91 @@ final class PropertyVisitorSpec: QuickSpec {
...
       }
 
+      context("when the property is an undefined constant property") {
+        let content = """
+        let foo: Foo
+        """
+
+        it("has expected paradigm") {
+          try sut.walkContent(content)
+          expect(sut.properties) == [
```
fdiaz, line 268: Maybe writing this as `sut.properties.first?.paradigm` and checking on the specific property might be more robust and resilient to refactors?

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
fdiaz: @bachand did we try something like `.withEqual(nil)` on the node?

```swift
/// Returns a copy of the receiver with its `equal` replaced.
  /// - param newChild: The new `equal` to replace the node's
  ///                   current `equal`, if present.
  public func withEqual(
    _ newChild: TokenSyntax?) -> InitializerClauseSyntax {
    let raw = newChild?.raw ?? RawSyntax.missingToken(TokenKind.equal)
    let newData = data.replacingChild(raw, at: Cursor.equal)
    return InitializerClauseSyntax(newData)
  }
```

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -135,6 +135,40 @@ public final class PropertyVisitor: SyntaxVisitor {
     return typeSyntax.typeDescription
   }
 
+  private func findParamadigm(from node: VariableDeclSyntax) -> PropertyInfo.Paradigm {
```
fdiaz, line 138: nit: findParamadigm -> findParadigm

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
fdiaz: Yup, that works https://github.com/fdiaz/SwiftInspector/pull/111

Sources/SwiftInspectorVisitors/Tests/PropertyInfoSpec.swift
bachand: Since there are a number of different permutations for these keys it's not completely clear to me how to write a simple helper. Can you help me understand what you're thinking? @fdiaz

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
bachand: I liked where we landed in that PR :)

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
bachand: https://github.com/fdiaz/SwiftInspector/pull/112/commits/095be15822fcf16b1aa1b86d7817f6ea353480aa

PR #111, opened 2021-09-16
Remove trivia from paradigm

fdiaz:
Acting on [this piece of feedback](https://github.com/fdiaz/SwiftInspector/pull/110#discussion_r710339525) I gave to @bachand PR

This simplifies the paradigm and only gives us the information about the type rather than the equals sign and the braces sign for computed properties.

cc @dfed

### DIFF
--- a/Sources/SwiftInspectorVisitors/PropertyInfo.swift
+++ b/Sources/SwiftInspectorVisitors/PropertyInfo.swift
@@ -74,8 +74,9 @@ extension PropertyInfo {
     /// - Parameter initializerDescription: A source-accurate description of the initializer.
     case definedVariable(_ initializerDescription: String)
     /// A computed `var` property.
-    /// - Parameter codeBlockDesciption: A source-accurate description of the code block which computes the value.
-    case computedVariable(_ codeBlockDesciption: String)
+    /// - Parameter codeBlockDescription: A source-accurate description of the code block which computes the value
+    /// - Important: The code block description does not include the opening/closing brackets
+    case computedVariable(_ codeBlockDescription: String)
 
     // MARK: Lifecycle
 
--- a/Sources/SwiftInspectorVisitors/PropertyInfo.swift
+++ b/Sources/SwiftInspectorVisitors/PropertyInfo.swift
@@ -67,15 +67,17 @@ extension PropertyInfo {
     case undefinedConstant
     /// A `let` property with an `=`.
     /// - Parameter initializerDescription: A source-accurate description of the initializer.
+    /// - Important: The initializer description does not include the equal sign.
     case definedConstant(_ initializerDescription: String)
     /// A  `var` property with no `=`.
     case undefinedVariable
     /// A `var ` property with an `=`.
     /// - Parameter initializerDescription: A source-accurate description of the initializer.
+    /// - Important: The initializer description does not include the equal sign.
     case definedVariable(_ initializerDescription: String)
     /// A computed `var` property.
     /// - Parameter codeBlockDescription: A source-accurate description of the code block which computes the value
-    /// - Important: The code block description does not include the opening/closing brackets
+    /// - Important: The code block description does not include the opening/closing braces.
     case computedVariable(_ codeBlockDescription: String)
 
     // MARK: Lifecycle
--- a/Sources/SwiftInspectorVisitors/PropertyInfo.swift
+++ b/Sources/SwiftInspectorVisitors/PropertyInfo.swift
@@ -14,7 +14,7 @@ public struct PropertyInfo: Codable, Hashable, CustomDebugStringConvertible {
   public let paradigm: Paradigm
 
   public var debugDescription: String {
-    "\(modifiers.rawValue) \(name) \(typeDescription?.asSource ?? "")"
+    "\(modifiers.rawValue) \(name) \(typeDescription?.asSource ?? "") \(paradigm)"
   }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/SwiftInspectorVisitors/PropertyInfo.swift
```diff
@@ -14,7 +14,7 @@ public struct PropertyInfo: Codable, Hashable, CustomDebugStringConvertible {
   public let paradigm: Paradigm
 
   public var debugDescription: String {
-    "\(modifiers.rawValue) \(name) \(typeDescription?.asSource ?? "")"
+    "\(modifiers.rawValue) \(name) \(typeDescription?.asSource ?? "") \(paradigm)"
```
fdiaz, line 17: We missed adding this to the debug description, so when tests were failing they were not showing the paradigm.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -158,7 +158,7 @@ public final class PropertyVisitor: SyntaxVisitor {
   }
 
   private func findInitializerDescription(from node: PatternBindingListSyntax) -> String? {
-    let initializerClauseSyntaxes = node.compactMap { $0.initializer }
+    let initializerClauseSyntaxes = node.compactMap { $0.initializer?.withEqual(nil) }
```
dfed, line 161: well this is a nifty trick!

dfed (approved):
this is great!

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -32,7 +32,7 @@ public final class PropertyVisitor: SyntaxVisitor {
 
   public override func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
     let modifier = findModifiers(from: node)
-    let paradigm = findParamadigm(from: node)
+    let paradigm = findParadigm(from: node)
```
bachand, line 35: Thanks

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -158,15 +158,25 @@ public final class PropertyVisitor: SyntaxVisitor {
...
+    let initializerClauseSyntaxes = node.compactMap { $0.initializer?.withEqual(nil) }
     assert(initializerClauseSyntaxes.count <= 1, "A property should have at most one initializer.")
     return initializerClauseSyntaxes.first?.description
   }
 
   private func findCodeBlockDescription(from node: PatternBindingListSyntax) -> String? {
-    let accessors = node.compactMap { $0.accessor }
-    assert(accessors.count <= 1, "A property should have at most one accessor.")
-    return accessors.first?.description
+    guard
```
bachand, line 167: @fdiaz and I spoke and felt that if we are removing the `=` we should also remove the `{` and `}` from computed properties. In the future we could make a sub-visitor to clean up some of this logic.

fdiaz:
@dfed PTAL, we talked with @bachand about the fact that we're removing the trivia for some things but not other, so we decided to also remove the brackets for computed properties since we already know these need this information, TAL at https://github.com/fdiaz/SwiftInspector/commit/1a9ce21e1d2607dbdac48d69bde89841037cee2f

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -158,15 +158,25 @@ public final class PropertyVisitor: SyntaxVisitor {
...
+      let patternBindingSyntax = node.children.first?.as(PatternBindingSyntax.self),
+      let codeBlockSyntax = patternBindingSyntax.children.compactMap({ $0.as(CodeBlockSyntax.self) }).first
+    else {
+      return nil
+    }
+    guard let codeBlockList = codeBlockSyntax.children
+            .compactMap ({ $0.as(CodeBlockItemListSyntax.self) })
```
bachand, line 174:
```suggestion
            .compactMap({ $0.as(CodeBlockItemListSyntax.self) })
```

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -373,7 +373,7 @@ final class PropertyVisitorSpec: QuickSpec {
               name: "foo",
               typeDescription: nil,
               modifiers: [.internal, .instance],
-              paradigm: .computedVariable("{ Foo() }"))
+              paradigm: .computedVariable("Foo()"))
```
bachand, line 376: This is nice symmetry, since the `computedVariable` case indicates that we will have `{` and `}`.

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -158,15 +158,25 @@ public final class PropertyVisitor: SyntaxVisitor {
...
     return initializerClauseSyntaxes.first?.description
   }
 
   private func findCodeBlockDescription(from node: PatternBindingListSyntax) -> String? {
-    let accessors = node.compactMap { $0.accessor }
-    assert(accessors.count <= 1, "A property should have at most one accessor.")
-    return accessors.first?.description
+    guard
+      let patternBindingSyntax = node.children.first?.as(PatternBindingSyntax.self),
+      let codeBlockSyntax = patternBindingSyntax.children.compactMap({ $0.as(CodeBlockSyntax.self) }).first
```
bachand, line 169: If we wanted we could assert here that there are only one of these nodes, though I feel like that's optional. We may want to come up with a general philosophy of when we think assertions are necessary when parsing nodes like this.

bachand (approved):
I like this improvement @fdiaz. Let's update the title and description of this PR before we merge.

dfed:
Second commit LGTM but I think we might want to update the doc comments on the cases for `Paradigm` that we're modifying here

Sources/SwiftInspectorVisitors/PropertyVisitor.swift
bachand: https://github.com/fdiaz/SwiftInspector/pull/112/commits/1bbfbc090e82922701ff2828c8c6f87342ab1de2

PR #112, opened 2021-09-16
Addresses feedback on recent PRs related to collecting more property information

bachand:
- I address this feedback from Francisco in the first commit: https://github.com/fdiaz/SwiftInspector/pull/110#discussion_r710336714
- I create a sub-visitor to simplify the code we need to write to extract the initializer and code block descriptions: https://github.com/fdiaz/SwiftInspector/pull/111#discussion_r710381745

@fdiaz cc: @dfed

### DIFF
--- a/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
@@ -136,44 +136,28 @@ public final class PropertyVisitor: SyntaxVisitor {
   }
 
   private func findParadigm(from node: VariableDeclSyntax) -> PropertyInfo.Paradigm {
+    let patternBindingListVisitor = PatternBindingListVisitor()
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -157,25 +160,6 @@ public final class PropertyVisitor: SyntaxVisitor {
     }
   }
 
-  private func findInitializerDescription(from node: PatternBindingListSyntax) -> String? {
-    let initializerClauseSyntaxes = node.compactMap { $0.initializer?.withEqual(nil) }
-    assert(initializerClauseSyntaxes.count <= 1, "A property should have at most one initializer.")
```
bachand, line 162: I think it's OK to avoid having these types of assertions.

fdiaz (approved):

PR #113, opened 2021-09-16
Update `PropertyInfo.Paradigm` to handle properties in protocols

bachand:
Best reviewed in Files Changed view.

This PR builds on https://github.com/fdiaz/SwiftInspector/pull/110 to add a paradigm for properties in protocols.

We start by documenting the current behavior.

Then we add ways to strongly encode these cases into `PropertyInfo.Paradigm`.

- `var foo: Foo { get }` -> `PropertyInfo.Paradigm.protocolGetter`
- `var foo: Foo { get set }` -> `PropertyInfo.Paradigm.protocolGetterAndSetter`

`var foo: Foo { set }` is not valid Swift code.

@fdiaz cc: @dfed

### DIFF
--- a/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
@@ -148,13 +148,17 @@ public final class PropertyVisitor: SyntaxVisitor {
         return .undefinedConstant
       }
     case .variable:
-      if let initializerDescription = patternBindingListVisitor.initializerDescription {
+      let initializerDescription = patternBindingListVisitor.initializerDescription
+      let codeBlockDescription = patternBindingListVisitor.codeBlockDescription
+      let protocolRequirement = patternBindingListVisitor.protocolRequirement
+
+      if let initializerDescription = initializerDescription {
         return .definedVariable(initializerDescription)
       }
-      else if let codeBlockDescription = patternBindingListVisitor.codeBlockDescription {
+      else if let codeBlockDescription = codeBlockDescription {
         return .computedVariable(codeBlockDescription)
       }
-      else if let protocolRequirement = patternBindingListVisitor.protocolRequirement {
+      else if let protocolRequirement = protocolRequirement {
         switch protocolRequirement {
         case .gettable: return .protocolGetter
         }
--- a/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/PropertyVisitor.swift
@@ -154,6 +154,11 @@ public final class PropertyVisitor: SyntaxVisitor {
       else if let codeBlockDescription = patternBindingListVisitor.codeBlockDescription {
         return .computedVariable(codeBlockDescription)
       }
+      else if let protocolRequirement = patternBindingListVisitor.protocolRequirement {
+        switch protocolRequirement {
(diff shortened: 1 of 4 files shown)


Sources/SwiftInspectorVisitors/PropertyVisitor.swift
```diff
@@ -176,10 +191,15 @@ public final class PropertyVisitor: SyntaxVisitor {
 
 private final class PatternBindingListVisitor: SyntaxVisitor {
 
+  enum ProtocolRequirement {
+    case gettable
```
bachand, line 195: Eventually we would add `gettableAndSettable`.

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -336,6 +336,31 @@ final class PropertyVisitorSpec: QuickSpec {
         }
       }
 
+      context("when the property is part of a protocol definition") {
+        context("and is only gettable") {
+          let content = """
+          var foo: Foo { get }
```
fdiaz, line 342: I think you want this example to be inside a protocol, if not this won't compile. In that case the property will be indeed inside a variable declaration.

```swift
protocol Some {
  var foo: Foo { get }
}
```

fdiaz (commented):
I think we might have an issue here with the test here and the example in the PR description.

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
bachand: I tried this at first. I then realized that we skip analyzing protocols.

image

We could invoke this visitor within a parent visitor that visits protocols.

Do you think we should do that? I expect we will end up with the same result though it could be nice to have a more realistic test.

fdiaz: Oh yeah that's a good point, we should get this inside of a protocol instead and visit through the protocol instead of directly here. I don't think we'd get the same result actually, because this code as is doesn't compile so it's giving us incorrect node information (you can see that the nodes are Unknown in your screenshot on top).

bachand: @fdiaz I made the change and we get the same result. https://github.com/fdiaz/SwiftInspector/commit/f079eac0af6e7e70e90d631b5e34af240ced918c, https://github.com/fdiaz/SwiftInspector/commit/889c6d335e581b53302032fd004fac8125e0d116

bachand:
I'm still very perplexed on Code Coverage. I made a new PR (to see if somehow CodeSense is corrupted on this PR) and we see a similar result. https://github.com/fdiaz/SwiftInspector/pull/114

bachand:
The issue with code coverage was that I had converted one of the tests from `it` to `fit` and checked in that change.

bachand:
@fdiaz this PR is now ready for review!

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
bachand: It works now!

fdiaz (approved):
Good stuff!

PR #108, opened 2021-09-11
Get SwiftInspector working with Xcode 13

dfed:
PR to get SwiftInspector to work with Swift 5.5

PR #116, opened 2021-09-24
Use exact tag for Swift 5.5

dfed:
It's out! Let's use it.

### DIFF
--- a/Package@swift-5.5.swift
+++ b/Package@swift-5.5.swift
@@ -16,7 +16,7 @@ let package = Package(
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.3.1")),
-    .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", .branch("release/5.5")),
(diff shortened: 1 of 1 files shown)


bachand:
Do we do the special packaging that allows this project to work across Xcode versions at this layer or a layer above? I am asking because I want to learn if changing the version requires any update to our packaging. I forget the exact name for what we do so hopefully that makes sense.

bachand (approved):

bachand:
Or maybe we only updating the packaging for Xcode versions?

fdiaz:
> Do we do the special packaging that allows this project to work across Xcode versions at this layer or a layer above? 

I'm not entirely sure what you mean by packaging. If you're talking about creating the binary, that's already independent of Xcode version and was done here https://github.com/fdiaz/SwiftInspector/pull/85

If you're referring to the fact that we can open SwiftInspector with versions of Xcode that contain different Swift versions, then this PR is the one that did it https://github.com/fdiaz/SwiftInspector/pull/108

The current PR is only updating what we did on https://github.com/fdiaz/SwiftInspector/pull/108 to now use the release version of SwiftSyntax instead of a branch.

bachand:
I was referring to https://github.com/fdiaz/SwiftInspector/pull/85. Thank you!

PR #118, opened 2021-09-27
Update dependencies

fdiaz:
Just updating deps to their latest versions.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,10 +15,10 @@ let package = Package(
     .library(name: "SwiftInspectorTestHelpers", targets: ["SwiftInspectorTestHelpers"]),
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.3.1")),
+    .package(url: "https://github.com/apple/swift-argument-parser", from: "1.0.0"),
     .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", .exact("0.50400.0")),
     .package(url: "https://github.com/Quick/Nimble.git", .upToNextMajor(from: "9.0.1")),
-    .package(url: "https://github.com/Quick/Quick.git", .upToNextMajor(from: "3.0.0")),
+    .package(url: "https://github.com/Quick/Quick.git", .upToNextMajor(from: "4.0.0")),
   ],
   targets: [
     .executableTarget(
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -15,10 +15,10 @@ let package = Package(
     .library(name: "SwiftInspectorTestHelpers", targets: ["SwiftInspectorTestHelpers"]),
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.3.1")),
+    .package(url: "https://github.com/apple/swift-argument-parser", from: "1.0.0"),
```
dfed, line 18: Don't we want:
```suggestion
    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMajor(from: "1.0.0")),
```

Package@swift-5.5.swift
```diff
@@ -15,10 +15,10 @@ let package = Package(
     .library(name: "SwiftInspectorTestHelpers", targets: ["SwiftInspectorTestHelpers"]),
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.3.1")),
+    .package(url: "https://github.com/apple/swift-argument-parser", from: "1.0.0"),
```
dfed, line 18: Same q here:
```suggestion
    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMajor(from: "1.0.0")),
```

dfed (approved):
LGTM with some blocking comments

Package.swift
fdiaz: I think this is what we want though? This is what the swift argument parser repo recommends too:

```
// Create a package dependency that uses the version requirement, starting with the given minimum version,
/// going up to the next major version.
///
/// This is the recommended way to specify a remote package dependency.
/// It allows you to specify the minimum version you require, allows updates that include bug fixes
/// and backward-compatible feature updates, but requires you to explicitly update to a new major version of the dependency.
/// This approach provides the maximum flexibility on which version to use,
/// while making sure you don't update to a version with breaking changes,
/// and helps to prevent conflicts in your dependency graph.
///
/// The following example allows the Swift Package Manager to select a version
/// like a  `1.2.3`, `1.2.4`, or `1.3.0`, but not `2.0.0`.
///
///    .package(url: "https://example.com/example-package.git", from: "1.2.3"),
///
/// - Parameters:
///     - name: The name of the package, or nil to deduce it from the URL.
///     - url: The valid Git URL of the package.
///     - version: The minimum version requirement.
static func package(url: String, from version: Version) -> Package.Dependency
```

lmk what you think

dfed: @fdiaz huh. Sounds like my suggestion and what you wrote are semantically equivalent then. I like what I wrote better because it's more clear, but my suggestions are no longer blocking.

PR #120, opened 2021-11-04
Make TypeDescription capable of understanding all types

dfed:
As title.

**Best reviewed commit-by-commit**

### DIFF
--- a/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
@@ -126,6 +126,21 @@ final class TypeDescriptionSpec: QuickSpec {
     }
     """.data(using: .utf8)!
 
+  let metatypeTestCase = TypeDescription.metatype(
+    .simple(name: "Foo"),
+    isType: false)
+  let metatypeTestCaseData = """
+    {
+      "caseDescription": "metatype",
+      "typeDescription": {
+        "caseDescription": "simple",
+        "text": "Foo",
+        "typeDescriptions": []
+      },
+      "isType": false
+    }
+    """.data(using: .utf8)!
+
   let someTestCase = TypeDescription.some(
     .simple(name: "Foo"))
   let someTestCaseData = """
@@ -293,6 +308,12 @@ final class TypeDescriptionSpec: QuickSpec {
         }
       }
 
+      context("that represents a metatype") {
+        it("decodes the encoded type description") {
+          expect(try decoder.decode(TypeDescription.self, from: self.metatypeTestCaseData)) == self.metatypeTestCase
+        }
+      }
+
       context("that represents a some-restricted type") {
         it("decodes the encoded type description") {
           expect(try decoder.decode(TypeDescription.self, from: self.someTestCaseData)) == self.someTestCase
@@ -381,6 +402,12 @@ final class TypeDescriptionSpec: QuickSpec {
       context("utilizing a composition type") {
         it("successfully decodes the data") {
           expect(try decoder.decode(TypeDescription.self, from: try encoder.encode(self.compositionTestCase))) == self.compositionTestCase
+        }
+      }
+
+      context("utilizing a metatype") {
+        it("successfully decodes the data") {
+          expect(try decoder.decode(TypeDescription.self, from: try encoder.encode(self.metatypeTestCase))) == self.metatypeTestCase
         }
       }
 
@@ -600,6 +627,47 @@ final class TypeDescriptionSpec: QuickSpec {
 
         it("Finds the type") {
           expect(visitor?.implictlyUnwrappedOptionalTypeIdentifier?.asSource) == "Int!"
+        }
+      }
+
+      context("when called on a TypeSyntax node representing a MetatypeTypeSyntax") {
+        final class MetatypeTypeSyntaxVisitor: SyntaxVisitor {
+          var metatypeTypeIdentifier: TypeDescription?
+          override func visit(_ node: MetatypeTypeSyntax) -> SyntaxVisitorContinueKind {
+            metatypeTypeIdentifier = TypeSyntax(node).typeDescription
+            return .skipChildren
+          }
+        }
+
+        var visitor: MetatypeTypeSyntaxVisitor!
+        context("that represents a type") {
+          beforeEach {
+            let content = """
+            let metatype: Int.Type
+            """
+
+            visitor = MetatypeTypeSyntaxVisitor()
+            try? visitor.walkContent(content)
+          }
+
+          it("Finds the type") {
+            expect(visitor.metatypeTypeIdentifier?.asSource) == "Int.Type"
+          }
+        }
+
+        context("that represents a protocol") {
+          beforeEach {
+            let content = """
+            let metatype: Equatable.Protocol
+            """
+
+            visitor = MetatypeTypeSyntaxVisitor()
+            try? visitor.walkContent(content)
+          }
+
+          it("Finds the type") {
+            expect(visitor.metatypeTypeIdentifier?.asSource) == "Equatable.Protocol"
+          }
         }
       }
 
--- a/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
@@ -126,6 +126,23 @@ final class TypeDescriptionSpec: QuickSpec {
     }
     """.data(using: .utf8)!
 
+  let attributedTestCase = TypeDescription.attributed(
+    .simple(name: "Foo"),
+    specifier: "inout",
+    attributes: ["autoclosure"])
+  let attributedTestCaseData = """
+    {
+      "caseDescription": "attributed",
+      "typeDescription": {
+        "caseDescription": "simple",
+        "text": "Foo",
+        "typeDescriptions": []
+      },
+      "specifier": "inout",
+      "attributes": ["autoclosure"],
+    }
+    """.data(using: .utf8)!
+
   let arrayTestCase = TypeDescription.array(element: .simple(name: "Foo"))
   let arrayTestCaseData = """
     {
@@ -263,6 +280,12 @@ final class TypeDescriptionSpec: QuickSpec {
         }
       }
 
+      context("that represents an attributed type") {
+        it("decodes the encoded type description") {
+          expect(try decoder.decode(TypeDescription.self, from: self.attributedTestCaseData)) == self.attributedTestCase
+        }
+      }
+
       context("that represents an array type") {
         it("decodes the encoded type description") {
           expect(try decoder.decode(TypeDescription.self, from: self.arrayTestCaseData)) == self.arrayTestCase
@@ -339,6 +362,12 @@ final class TypeDescriptionSpec: QuickSpec {
       context("utilizing a composition type") {
         it("successfully decodes the data") {
           expect(try decoder.decode(TypeDescription.self, from: try encoder.encode(self.compositionTestCase))) == self.compositionTestCase
+        }
+      }
+
+      context("utilizing an attributed type") {
+        it("successfully decodes the data") {
+          expect(try decoder.decode(TypeDescription.self, from: try encoder.encode(self.attributedTestCase))) == self.attributedTestCase
         }
       }
 
@@ -546,6 +575,63 @@ final class TypeDescriptionSpec: QuickSpec {
 
         it("Finds the type") {
           expect(visitor?.implictlyUnwrappedOptionalTypeIdentifier?.asSource) == "Int!"
+        }
+      }
+
+      context("when called on a TypeSyntax node representing an AttributedTypeSyntax") {
+        final class AttributedTypeSyntaxVisitor: SyntaxVisitor {
+          var attributedTypeIdentifier: TypeDescription?
+          override func visit(_ node: AttributedTypeSyntax) -> SyntaxVisitorContinueKind {
+            attributedTypeIdentifier = TypeSyntax(node).typeDescription
+            return .skipChildren
+          }
+        }
+
+        var visitor: AttributedTypeSyntaxVisitor!
+        context("with a specifier") {
+          beforeEach {
+            let content = """
+            inout Int
+            """
+
+            visitor = AttributedTypeSyntaxVisitor()
+            try? visitor.walkContent(content)
+          }
+
+          it("Finds the type") {
+            expect(visitor.attributedTypeIdentifier?.asSource) == "inout Int"
+          }
+        }
+
+        context("with attributes") {
+          beforeEach {
+            let content = """
+            @autoclosure () -> Void
+            """
+
+            visitor = AttributedTypeSyntaxVisitor()
+            try? visitor.walkContent(content)
+          }
+
+          it("Finds the type") {
+            expect(visitor.attributedTypeIdentifier?.asSource) == "@autoclosure () -> Void"
+          }
+
+          context("with a specifier and attributes") {
+            beforeEach {
+              let content = """
+              // This code doesn't compile but it can be parsed.
+              inout @autoclosure () -> Void
+              """
+
+              visitor = AttributedTypeSyntaxVisitor()
+              try? visitor.walkContent(content)
+            }
+
+            it("Finds the type") {
+              expect(visitor.attributedTypeIdentifier?.asSource) == "inout @autoclosure () -> Void"
+            }
+          }
         }
       }
 
--- a/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
+++ b/Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
@@ -126,6 +126,19 @@ final class TypeDescriptionSpec: QuickSpec {
     }
     """.data(using: .utf8)!
 
+  let someTestCase = TypeDescription.some(
+    .simple(name: "Foo"))
+  let someTestCaseData = """
+    {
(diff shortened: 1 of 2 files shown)


Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -267,9 +303,8 @@ public enum TypeDescription: Codable, Hashable {
 extension TypeSyntax {
 
   /// Returns the type description for the receiver.
-  /// - Warning: Do not call on a type syntax node of type `ClassRestrictionTypeSyntax`,
```
dfed, line 270: this warning got out-of-date! Turns out including the same information twice can lead to inconsistencies 😬

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -34,8 +34,10 @@ public enum TypeDescription: Codable, Hashable {
   indirect case composition([TypeDescription])
   /// An optional type. e.g. Int?
   indirect case optional(TypeDescription)
-  /// An implicitly unwrapped optional type. eg. Int!
+  /// An implicitly unwrapped optional type. e.g. Int!
```
dfed, line 37: unrelated change but noticed this typo while I was here

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -104,6 +105,24 @@ public enum TypeDescription: Codable, Hashable {
...
+          .map { "@\($0)" }
+          .joined(separator: " ")
+      }
+      switch (specifier, attributes) {
+      case let (.some(specifier), .none):
+        return "\(specifier) \(type.asSource)"
+      case let (.none, .some(attributes)):
+        return "\(attributesFromList(attributes)) \(type.asSource)"
+      case let (.some(specifier), .some(attributes)):
+        // This case likely represents an error.
```
dfed, line 118: I'm not _convinced_ of this but I've been having trouble coming up with legitimate Swift syntax where we have both a specifier and attributes

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -104,6 +105,24 @@ public enum TypeDescription: Codable, Hashable {
       } else {
         return "\(parentType.asSource).\(name)<\(generics.map { $0.asSource }.joined(separator: ", "))>"
       }
+    case let .attributed(type, specifier: specifier, attributes: attributes):
+      func attributesFromList(_ attributes: [String]) -> String {
+        attributes
+          .map { "@\($0)" }
```
dfed, line 109: Per [the Swift reference manual](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#grammar_attributes), the attribute grammar always starts with an `@`:
image

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -104,6 +105,24 @@ public enum TypeDescription: Codable, Hashable {
...
+      switch (specifier, attributes) {
+      case let (.some(specifier), .none):
+        return "\(specifier) \(type.asSource)"
+      case let (.none, .some(attributes)):
+        return "\(attributesFromList(attributes)) \(type.asSource)"
+      case let (.some(specifier), .some(attributes)):
+        // This case likely represents an error.
+        return "\(specifier) \(attributesFromList(attributes)) \(type.asSource)"
+      case (.none, .none):
+        // This case likely represents an error.
```
dfed, line 121: This case, however, I'm pretty convinced is an error. I don't know how we'd end up as an attributed type without either a specifier or attributes

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -104,6 +105,24 @@ public enum TypeDescription: Codable, Hashable {
...
+          .joined(separator: " ")
+      }
+      switch (specifier, attributes) {
+      case let (.some(specifier), .none):
+        return "\(specifier) \(type.asSource)"
+      case let (.none, .some(attributes)):
+        return "\(attributesFromList(attributes)) \(type.asSource)"
+      case let (.some(specifier), .some(attributes)):
+        // This case likely represents an error.
+        return "\(specifier) \(attributesFromList(attributes)) \(type.asSource)"
```
dfed, line 119: I'm reasonably sure this ordering is both correct and stable.

I was playing around with https://swift-ast-explorer.com/ and `inout @autoclosure () -> Void` parsed as an `AttributedType` but `@autoclosure inout () -> Void` parsed into `UnknownType`... so I'm pretty sure the specifier has to come first.

That said, [the Swift reference manual](https://docs.swift.org/swift-book/ReferenceManual/Types.html#grammar_type) shows the specifier (specifically `inout`) coming after attributes:
image

Not really sure what to do with this one. Though given that I haven't yet come up with a compiling statement where both the specifier and attributes are present I'm not super worried about it.

If you can come up with a compiling type declaration that has both a specifier and attributes, I'd love to see it.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -106,9 +99,11 @@ public enum TypeDescription: Codable, Hashable {
       } else {
         return "\(parentType.asSource).\(name)<\(generics.map { $0.asSource }.joined(separator: ", "))>"
       }
+    case let .metatype(type, isType):
+      return "\(type.asSource).\(isType ? "Type" : "Protocol")"
```
dfed, line 103: While this looks like a hack, it's inline with what [the Swift reference manual](https://docs.swift.org/swift-book/ReferenceManual/Types.html#grammar_metatype-type) says are our possible options here:
image

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -76,15 +78,6 @@ public enum TypeDescription: Codable, Hashable {
     .nested(name: name, parentType: parentType, generics: [])
   }
 
-  /*
-   * Note that we do not yet support the following syntax types:
-   * MetatypeTypeSyntax
-   * UnknownTypeSyntax
-   *
-   * We will likely need to add these types at some point in the future.
-   * We’ll Get There™
```
dfed, line 83: We Got There™!

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -314,8 +322,6 @@ public enum TypeDescription: Codable, Hashable {
 extension TypeSyntax {
 
   /// Returns the type description for the receiver.
-  /// - Warning: Do not call on a type syntax node of type `MetatypeTypeSyntax`
-  ///            or `UnknownTypeSyntax`
```
dfed: We _do_ handle `UnknownTypeSyntax` in our `else` statement. There's no real way to handle `UnknownTypeSyntax` since the type is unknown.

As far as I can tell, `UnknownTypeSyntax` is only utilized when the code will not compile. What we do in this case is grab the `description`, which is basically our only option.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -379,7 +390,7 @@ extension TypeSyntax {
         returnType: typeIdentifier.returnType.typeDescription)
 
     } else {
-      assertionFailureOrPostNotification("TypeSyntax of unexpected type. Defaulting to `description`.")
```
dfed, line 325: no longer unexpected!

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Note that because I believe this is an error I haven't been able to write a test case that covers this case. Same goes for the case below.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Note that because I believe this is an error I haven't been able to write a test case that covers this case.

Sources/SwiftInspectorVisitors/Tests/TypeDescriptionSpec.swift
```diff
@@ -549,6 +630,128 @@ final class TypeDescriptionSpec: QuickSpec {
...
+          }
+
+          it("Finds the type") {
+            expect(visitor.attributedTypeIdentifier?.asSource) == "@autoclosure () -> Void"
+          }
+
+          context("with a specifier and attributes") {
+            beforeEach {
+              let content = """
+              // This code doesn't compile but it can be parsed.
```
dfed, line 740: as mentioned elsewhere I do not know if there exists compiling syntax that declares an attributed type with both a specifier and attributes

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Note that because I believe this is an error I haven't been able to write a test case that covers this case.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Note that because I believe this is an error I haven't been able to write a test case that covers this case.

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -104,6 +105,24 @@ public enum TypeDescription: Codable, Hashable {
       } else {
         return "\(parentType.asSource).\(name)<\(generics.map { $0.asSource }.joined(separator: ", "))>"
       }
+    case let .attributed(type, specifier: specifier, attributes: attributes):
```
bachand: Can specifiers only exist on attributed types?

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: Are property wrappers "attributes" in the grammar?

bachand: Newer versions of Swift support property wrappers on function arguments https://github.com/apple/swift-evolution/blob/main/proposals/0293-extend-property-wrappers-to-function-and-closure-parameters.md.

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: I wondered if we could do something with property wrappers on function parameters though they don't support `inout` at the moment https://github.com/apple/swift-evolution/blob/main/proposals/0293-extend-property-wrappers-to-function-and-closure-parameters.md#support-inout-in-wrapped-function-parameters

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: Let's add more context to this comment, i.e. that we can't figure out compiling code.

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: Let's assert?

Sources/SwiftInspectorVisitors/TypeDescription.swift
bachand: Let's add a comment that we are aware this conflicts with the grammar and explain why we are writing this the way we are.

bachand (approved):
Hurray!

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: As far as I'm aware, this is correct.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I believe so, yes. But they'd be `attributes` on a `***DeclSyntax` type rather than an `AttributedTypeSyntax` since property wrappers are on the type declaration rather than the type reference.

We don't currently track attributes on type declarations. We’ll Get There™ (once we find a need for them)

dfed: The property wrappers on functions and closures look like these are attributes, but on the property rather than a type since the attribute comes before the property name. An example in that doc is `@Traceable value: Value`. If this wrapper looked like `value: @Traceable Value` then yes this would be an attribute on the type rather than the property.

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Per [my comment in the other thread](https://github.com/fdiaz/SwiftInspector/pull/120/files#r743379615) I don't think property wrappers or function parameters are attributes on the type, but rather are attributes on the property.

I've added a better block comment here however. af09cb1

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: Added an assert in two spots for good measure, and avoided surfacing the type we think is a bug at the parsing stage:  5b44aac3dc93f45ed9300f45752a194e975d58c8

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: addressed in af09cb1

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I did a little more digging. `ClosureCaptureItemSyntax` also has a `specifier`. I'm not yet sure what the options are for this syntax.

bachand: Screen Shot 2021-11-05 at 6 00 29 PM

I wonder if `weak` is a specifier?

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -116,6 +116,12 @@ public enum TypeDescription: Codable, Hashable {
         return "\(attributesFromList(attributes)) \(type.asSource)"
       case let (.some(specifier), .some(attributes)):
         // This case likely represents an error.
+        // We are unaware of type reference that compiles with both a specifier and attributes.
+        // The Swift reference manual specifies that attributes come before the specifier,
+        // however code that puts an attribute first does not parse as AttributedTypeSyntax.
+        // Only code where the specifier comes before the attribute parses as an AttributedTypeSyntax.
+        // As a result, we construct this source with the specifier first.
+        // Reference manual: https://docs.swift.org/swift-book/ReferenceManual/Types.html#grammar_type
```
bachand, line 124: Great!

Sources/SwiftInspectorVisitors/TypeDescription.swift
```diff
@@ -366,12 +367,19 @@ extension TypeSyntax {
...
-        attributes: typeIdentifier.attributes?.compactMap {
-          $0.as(AttributeSyntax.self)?.attributeName.text
-        })
+      let specifier = typeIdentifier.specifier?.text
+      let attributes = typeIdentifier.attributes?.compactMap {
+        $0.as(AttributeSyntax.self)?.attributeName.text
+      }
+      if specifier == nil && attributes == nil {
+        assertionFailureOrPostNotification("Encountered an attributed type with no attributes or specifiers")
+        return typeIdentifier.baseType.typeDescription
```
bachand, line 376: Hmm if we actually hit this code path it seems to indicate that we've found potentially valid syntax we didn't expect, right? I wonder if it makes sense to keep propagating `.attributed(...)` to ensure that we persist the data even if we may not correctly generate source from it today? Curious to learn more about this change.

bachand (approved):

Sources/SwiftInspectorVisitors/TypeDescription.swift
dfed: I think this would only happen if properties beyond `specifier` and `attributes` were added to this type.

We can keep propagating `.attributed` here. I changed to propagating the `baseType` because I figured our `asSource` would produce the same result and we should consolidate.

I'll make this change to keep propagating `.attributed` here, and in that commit I think I'll also remove the assert in `asSource` since we'll have asserted here when we hit the error case (and since the `asSource` code will now be expected if we've hit this assert). It feels better to assert here because we know we'll hit this code if we parse it, and who knows if someone will call `asSource` on a type.

dfed: 3a6712a

bachand: > I think this would only happen if properties beyond `specifier` and `attributes` were added to this type.

I hadn't considered this possibility.

Change LGTM.

dfed:
I don't understand how Codecov is seeing such a big diff in the `patch` but not the `project`:
image

I'm going to rebase `main` to see if I can kick this thing into working as I expect

dfed:
I think our Codecov patch calculation is somehow wrong. Running locally, I see:
image

And specifically for the file changed:
image

Compare to this file on `main`:
image

The lines that are missing are all lines where we no longer know how to hit the fallback due to an unexpected syntax type.

I'm going to merge through the failing check.

bachand:
The way I read the codecov results is that for the lines we've changed in this PR we have 92.85% code coverage (patch) but for the codebase overall we have 97.41% (project). I didn't do the math myself but it seems possible that we could end up in that situation. I wonder: do we care about codecov enforcing that the lines changed in each PR/patch hit a certain level of code coverage? Could we remove the patch check? cc: @fdiaz

fdiaz:
> do we care about codecov enforcing that the lines changed in each PR/patch hit a certain level of code coverage? Could we remove the patch check? 

I don't think so, yeah we can def remove the patch check if that's the case.

PR #122, opened 2022-01-24
Update README to be more precise

bachand:
Someone who finds this repository may not be as familiar as we are with the `SwiftSyntax` toolchain or the underling concepts. Let's make our project more approachable.


dfed (approved):
Solid improvement

PR #123, opened 2022-02-18
Remove need to use 'swift package generate-xcodeproj'

dfed:
Thank you to @qyang-nj for figuring out how to do this!


qyang-nj:
Nice! It looks good to me, but I don't have permission to approve.

fdiaz (approved):
Soooo cool! Thanks @qyang-nj you da best!

PR #125, opened 2022-02-22
Only test project coverage

dfed:
Inspired by https://github.com/fdiaz/SwiftInspector/pull/120#issuecomment-1048231079. I don't know if this is the right way to do this though, hence the draft PR


dfed:
Deferring to #126

PR #126, opened 2022-02-22
Remove patch from codecov.yml

fdiaz:
Based on the [cloudcov documentation](https://docs.codecov.com/docs/commit-status#patch-status) and what @bachand mentioned [here](https://github.com/fdiaz/SwiftInspector/pull/120#issuecomment-1048231079) we should remove the patch from the coverage enforcement.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,9 +1,5 @@
 coverage:
   status:
-    # PRs only
-    patch:
-      default:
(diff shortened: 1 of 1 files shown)


dfed (approved):
thank you!

bachand:
Thanks!

PR #130, opened 2022-04-29
Use DeclarationModifierVisitor in more places

dfed:
I realize we don't have modifiers on our function declaration info! So I pulled out our modifier-finding logic and reused it.


Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
```diff
@@ -42,6 +48,7 @@ fileprivate final class FunctionSignatureVisitor: SyntaxVisitor {
 }
 
 public struct FunctionDeclarationInfo: Codable, Hashable {
+  public let modifiers: [String]
```
dfed, line 51: Our `PropertyInfo` has a strongly typed `Modifiers` that isn't directly applicable here (since properties have modifiers which don't apply to functions like `private(set)`). We could create a new strongly typed `Modifiers` for use in function declarations, or we could just stick to a `String` array for now. I think I prefer the simplicity of working with Strings at this level, but maybe I'm open to being convinced that I should use a stronger type.

Sources/SwiftInspectorVisitors/ModifierDeclarationVisitor.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import SwiftSyntax
+
+public final class ModifierDeclarationVisitor: SyntaxVisitor {
+  public override func visit(_ node: DeclModifierSyntax) -> SyntaxVisitorContinueKind {
+    if
+      let leftParen = node.detailLeftParen,
+      let detail = node.detail,
+      let rightParen = node.detailRightParen
+    {
```
dfed, line 33: this logic was stolen/moved from PropertyVisitor

Sources/SwiftInspectorVisitors/DeclarationModifierVisitor.swift
```diff
@@ -29,8 +29,17 @@ final class DeclarationModifierVisitor: SyntaxVisitor {
 
   private(set) var modifiers = [String]()
 
-  override func visit(_ node: DeclModifierSyntax) -> SyntaxVisitorContinueKind {
-    modifiers.append(node.name.text)
+  public override func visit(_ node: DeclModifierSyntax) -> SyntaxVisitorContinueKind {
+    if
```
dfed, line 33: this logic was stolen/moved from PropertyVisitor

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
```diff
@@ -3,16 +3,44 @@ import SwiftSyntax
...
     let info = FunctionDeclarationInfo(
+      modifiers: modifiersVisitor.modifiers,
       name: name,
       arguments: functionSignatureVisitor.arguments,
       returnType: functionSignatureVisitor.returnType)
     functionDeclarations.append(info)
     return .skipChildren
   }
 
+  public override func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
```
dfed, line 24: this change could be pulled out into a separate PR, but I realized our `FunctionDeclarationVisitor` was likely visiting functions from nested classes, and we really don't want it doing that (I can write a test for this before merging but wanted to get this thought down before I forgot 😅)

dfed: might be good for me to pull this out into its own PR. I originally put this code here as a perf improvement and then realized that we likely have a bug here.

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: I'd prefer to use an array of `Modifier` similar to the PropertyInfo (we likely wanna move that to not be nested inside `PropertyInfo `). I don't see a modifier not being available to a function as problematic since this is not enums, but rather an `OptionSet` and not everything in an OptionSet IMO needs to be available to everything.

The nice thing about the Modifier is that we can have multiple things all at once like `static` and `public` in the same type, and if we want to move to a String from there its trivial, but doing it the other way around (from string -> single modifiers) requires us to parse the String again, which seems error prone.

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
dfed: gonna revert this here and instead fix this in https://github.com/fdiaz/SwiftInspector/pull/131

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
dfed: This is now a breaking change, and I've unified our usage of `Modifiers`. Let me know what you think!

fdiaz (approved):
Good stuff 👍

bachand:
Deferring to Francisco's review.

PR #131, opened 2022-04-29
Walk less of the syntax tree, and tighten up our contract for Properties and Functions visitors

dfed:
Today, our `NestableTypeVisitor` and `ExtensionVisitor` both walk the entire node tree to find properties and again to find functions. This leads to a performance hit (we're doing more than we have to!), and also led to a very real bug on `main`: our `FunctionsVisitor` attributes functions found in nested classes to the parent class type.

This PR resolves these issues by creating the `PropertyVisitor` and `FunctionVisitor` just-in-time when we encounter the expected declaration. It also tightens up the contract to make running both `PropertyVisitor` and `FunctionVisitor` over unexpected type declarations a contract violation.

### DIFF
--- a/Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
@@ -60,19 +60,13 @@ public final class ExtensionVisitor: SyntaxVisitor {
       declarationModifierVisitor.walk(modifiers)
     }
 
-    let propertyVisitor = PropertyVisitor()
-    propertyVisitor.walk(node.members)
-
-    let functionDeclarationVisitor = FunctionDeclarationVisitor()
-    functionDeclarationVisitor.walk(node)
-
     extensionInfo = ExtensionInfo(
       typeDescription: node.extendedType.typeDescription,
       inheritsFromTypes: typeInheritanceVisitor.inheritsFromTypes,
       genericRequirements: genericRequirementVisitor.genericRequirements,
       modifiers: .init(declarationModifierVisitor.modifiers),
-      properties: propertyVisitor.properties,
-      functionDeclarations: functionDeclarationVisitor.functionDeclarations)
+      properties: [],
+      functionDeclarations: [])
     return .visitChildren
   }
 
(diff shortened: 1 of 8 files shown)


Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
```diff
@@ -60,19 +60,13 @@ public final class ExtensionVisitor: SyntaxVisitor {
       declarationModifierVisitor.walk(modifiers)
     }
 
-    let propertyVisitor = PropertyVisitor()
-    propertyVisitor.walk(node.members)
```
dfed, line 64: another benefit of not walking the entire declaration node here is that our property visitor doesn't need to be hardened against accidentally reading properties from nested classes (doesn't hurt to leave that hardening but this approach still leaves us with fewer ways we can hurt ourselves)

Sources/SwiftInspectorVisitors/ExtensionVisitor.swift
```diff
@@ -133,9 +147,9 @@ public final class ExtensionVisitor: SyntaxVisitor {
 
 public struct ExtensionInfo: Codable, Hashable {
   public let typeDescription: TypeDescription
-  public private(set) var inheritsFromTypes: [TypeDescription]
-  public private(set) var genericRequirements: [GenericRequirement]
```
dfed, line 137: I don't know why these were previously mutable

Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
```diff
@@ -929,6 +930,8 @@ final class NestableTypeVisitorSpec: QuickSpec {
             $0.name == "FooEnum"
               && $0.inheritsFromTypes.map { $0.asSource } == []
               && $0.parentType?.asSource == nil
+              && $0.properties.isEmpty
+              && $0.functionDeclarations.isEmpty
```
dfed, line 934: this check fails on `main` today

Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
```diff
@@ -1002,6 +1005,12 @@ final class NestableTypeVisitorSpec: QuickSpec {
           expect(property?.name) == "someFooFooClassProperty"
         }
 
+        it("finds the function declared in FooEnum.FooFooClass") {
```
dfed, line 1008: we don't _need_ this test but it's good to have it – we already have a similar test for properties above

Sources/SwiftInspectorVisitors/Tests/PropertyVisitorSpec.swift
```diff
@@ -373,44 +374,80 @@ final class PropertyVisitorSpec: QuickSpec {
         }
       }
 
-      context("when there is a type declaration in content") {
```
dfed, line 376: this context is now considered API misuse.

Sources/SwiftInspectorVisitors/Tests/NestableTypeVisitorSpec.swift
fdiaz: We don't need to do this today, but when this test fails it's gonna be really annoying to debug. It will tell us "expected matching.count of 1, got XYZ.

That forces us to debug the test, which defeats the purpose of having a robust test suite.

I'd suggest refactoring these tests so when they fail they're trivial to understand why they failed, but we don't need to that in this PR.

fdiaz (approved):

bachand (commented):
Deferring to Francisco's review. Thanks Dan!

PR #133, opened 2022-06-02
Update dependencies and binary to support releasing with Swift 5.6

dfed:
As title. Also got our script capable of releasing from an M1 Mac. This change was inspired by work done by @fdiaz elsewhere.

I tested this by running unit tests in the Package.swift project, and by running `make release` and executing that binary. I did all of this on an Intel Mac.

cc @qyang-nj

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ xcode_toolchain ?= $(xcode_path)/Toolchains/XcodeDefault.xctoolchain/usr/lib/swi
 
 .PHONY: build
 build:
-	swift build -c release --arch arm64 --arch x86_64
+	xcrun swift build -c release --arch arm64 --arch x86_64
 
 .PHONY: install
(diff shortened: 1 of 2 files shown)


fdiaz (approved):

Makefile
```diff
@@ -5,7 +5,7 @@ xcode_toolchain ?= $(xcode_path)/Toolchains/XcodeDefault.xctoolchain/usr/lib/swi
 
 .PHONY: build
 build:
-	swift build -c release
+	swift build -c release --arch arm64 --arch x86_64
```
dfed, line 8: Let's make sure we're using Xcode's `swift` version, not the system's:
```suggestion
	xcrun swift build -c release --arch arm64 --arch x86_64
```

dfed:
Just rebased `main` to get CI working

PR #136, opened 2022-09-16
Add support for parameter names

fdiaz:
This PR adds support for parameter names in functions.

We were already supporting argument labels so adding this was fairly trivial.

### DIFF
--- a/Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
@@ -15,7 +15,7 @@ public final class FunctionDeclarationVisitor: SyntaxVisitor {
     let info = FunctionDeclarationInfo(
       modifiers: modifiersVisitor.modifiers,
       name: name,
-      arguments: functionSignatureVisitor.arguments,
+      parameters: functionSignatureVisitor.parameters,
       returnType: functionSignatureVisitor.returnType)
     functionDeclarations.append(info)
     return .skipChildren
@@ -52,13 +52,13 @@ public final class FunctionDeclarationVisitor: SyntaxVisitor {
 fileprivate final class FunctionSignatureVisitor: SyntaxVisitor {
   override func visit(_ node: FunctionParameterSyntax) -> SyntaxVisitorContinueKind {
     guard
-      let argumentLabelName = node.firstName?.text,
+      let firstMember = node.firstName?.text,
       let type = node.type?.typeDescription
     else { return .skipChildren }
 
-    let parameterName = node.secondName?.text
+    let secondMember = node.secondName?.text
 
-    appendArgument(argumentLabelName: argumentLabelName, parameterName: parameterName, type: type)
+    appendParameter(firstMember: firstMember, secondMember: secondMember, type: type)
     return .skipChildren
   }
   override func visit(_ node: ReturnClauseSyntax) -> SyntaxVisitorContinueKind {
@@ -66,30 +66,33 @@ fileprivate final class FunctionSignatureVisitor: SyntaxVisitor {
     return .skipChildren
   }
 
-  fileprivate func appendArgument(argumentLabelName: String, parameterName: String?, type: TypeDescription) {
-    var arguments = self.arguments ?? []
-    // By default, parameters use their parameter name as their argument label
-    let param = parameterName ?? argumentLabelName
-    arguments.append(.init(argumentLabelName: argumentLabelName, parameterName: param, type: type))
-    self.arguments = arguments
+  fileprivate func appendParameter(firstMember: String, secondMember: String?, type: TypeDescription) {
+    var parameters = self.parameters ?? []
+    // Each function parameter has both an argument label and a parameter name.
+    // The argument label is used when calling the function; each argument is written in the function call with its argument label before it.
+    // The parameter name is used in the implementation of the function. By default, parameters use their parameter name as their argument label.
+    let argumentLabel = firstMember
+    let parameterName = secondMember ?? firstMember
+    parameters.append(.init(argumentLabelName: argumentLabel, parameterName: parameterName, type: type))
+    self.parameters = parameters
   }
 
-  fileprivate var arguments: [FunctionDeclarationInfo.ArgumentInfo]?
+  fileprivate var parameters: [FunctionDeclarationInfo.ParameterInfo]?
   fileprivate var returnType: TypeDescription?
 }
 
 public struct FunctionDeclarationInfo: Codable, Hashable {
   public let modifiers: Modifiers
   public let name: String
-  public let arguments: [ArgumentInfo]?
+  public let parameters: [ParameterInfo]?
   public let returnType: TypeDescription?
 
   /// A convenience for creating a selector string that can be reference in Objective-C code.
   public var selectorName: String {
-    "\(name)(\((arguments ?? []).map { "\($0.argumentLabelName):" }.joined()))"
+    "\(name)(\((parameters ?? []).map { "\($0.argumentLabelName):" }.joined()))"
   }
 
-  public struct ArgumentInfo: Codable, Hashable {
+  public struct ParameterInfo: Codable, Hashable {
     public let argumentLabelName: String
     public let parameterName: String
     public let type: TypeDescription
--- a/Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
+++ b/Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
@@ -56,7 +56,9 @@ fileprivate final class FunctionSignatureVisitor: SyntaxVisitor {
       let type = node.type?.typeDescription
     else { return .skipChildren }
 
-    appendArgument(argumentLabelName: argumentLabelName, type: type)
+    let parameterName = node.secondName?.text
+
+    appendArgument(argumentLabelName: argumentLabelName, parameterName: parameterName, type: type)
     return .skipChildren
   }
   override func visit(_ node: ReturnClauseSyntax) -> SyntaxVisitorContinueKind {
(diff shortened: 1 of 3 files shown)


shepting (approved):
LGTM.

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
```diff
@@ -56,17 +56,21 @@ fileprivate final class FunctionSignatureVisitor: SyntaxVisitor {
...
     return .skipChildren
   }
 
-  fileprivate func appendArgument(argumentLabelName: String, type: TypeDescription) {
+  fileprivate func appendArgument(argumentLabelName: String, parameterName: String?, type: TypeDescription) {
     var arguments = self.arguments ?? []
-    arguments.append(.init(argumentLabelName: argumentLabelName, type: type))
+    // By default, parameters use their parameter name as their argument label
+    let param = parameterName ?? argumentLabelName
+    arguments.append(.init(argumentLabelName: argumentLabelName, parameterName: param, type: type))
```
dfed, line 73: I don't love the name `param`. I might write this as:
```suggestion
    // By default, parameters use their parameter name as their argument label
    let param = parameterName ?? argumentLabelName
    arguments.append(.init(
      argumentLabelName: argumentLabelName,
      parameterName: parameterName ?? argumentLabelName,
      type: type))
```
As an aside, the comment `By default, parameters use their parameter name as their argument label` doesn't make a whole lot of sense. The parameter name only falls back to the argument label if there is no explicit parameter name, right?

dfed (approved):
LGTM with a minor nit

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
bachand: I do think the comment `By default, parameters use their parameter name as their argument label` is correct per [this](https://docs.swift.org/swift-book/LanguageGuide/Functions.html#ID166)
Screen Shot 2022-09-16 at 2 33 03 PM

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
```diff
@@ -56,17 +56,21 @@ fileprivate final class FunctionSignatureVisitor: SyntaxVisitor {
...
+    appendArgument(argumentLabelName: argumentLabelName, parameterName: parameterName, type: type)
     return .skipChildren
   }
   override func visit(_ node: ReturnClauseSyntax) -> SyntaxVisitorContinueKind {
     returnType = node.returnType.typeDescription
     return .skipChildren
   }
 
-  fileprivate func appendArgument(argumentLabelName: String, type: TypeDescription) {
+  fileprivate func appendArgument(argumentLabelName: String, parameterName: String?, type: TypeDescription) {
```
bachand, line 69: Technically speaking if a function argument only has a parameter name and no argument label then the single name is the parameter name. The way we've written this function though the single name would be the argument label.

bachand: Can we rework this function so that the argument label name is optional and the parameter name is required?

Sources/SwiftInspectorVisitors/Tests/FunctionDeclarationVisitorSpec.swift
```diff
@@ -216,10 +222,14 @@ final class FunctionDeclarationVisitorSpec: QuickSpec {
         it("finds the correct number of parameters") {
           expect(self.sut.functionDeclarations.first?.arguments?.count) == 2
         }
-        it("finds the parameter labels") {
+        it("finds the argument labels") {
```
bachand, line 225: Can we add a test somewhere for a case where there is no argument label and only a parameter name?

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
```diff
@@ -87,6 +91,7 @@ public struct FunctionDeclarationInfo: Codable, Hashable {
 
   public struct ArgumentInfo: Codable, Hashable {
     public let argumentLabelName: String
+    public let parameterName: String
```
bachand, line 94: I was going to say that we should make the argument label optional but since Swift stresses both exist I think this is fine.
Screen Shot 2022-09-16 at 2 37 30 PM

bachand (approved):

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
dfed: That's helpful context! Within that context, the comment makes sense. Putting that comment on top of the assignment of `param` is what confused me: we're setting the parameter name in the next line, not the argument label. And we're setting the parameter name to be the same as the argument label, rather than setting the argument label to be the same as the parameter name which is what the comment suggests we should be doing.

Maybe what's confusing me more is the fact that we're setting the `parameterName` to be `argumentLabelName` if there is no explicit `parameterName`. In fact, doing this loses data.

This code would create identical output for the following scanned methods:
```
func doSomething(parameterName parameterName: Void)
func doSomething(parameterName: Void)
```

Ideally the output of our scans could be used to re-create the original code (or at least the method signatures on the scanned types). This implementation doesn't let us do that with confidence

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
dfed: Per my realization at the end of https://github.com/fdiaz/SwiftInspector/pull/136#discussion_r973431564 I think making the `parameterName` optional would be helpful.

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: > I don't love the name param. I might write this as:

I thought about doing that but wanted to take it out into its own line so I could add the code comment above. Since this is in the implementation and I _think_ param pretty clearly refers to "parameter" I think this is fine? But lemme know if you disagree!

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: Oh yeah, I think I got this the other way around. I disagree with making these optional though, that seems incorrect since we'd never really have a nil argument label.

Sources/SwiftInspectorVisitors/Tests/FunctionDeclarationVisitorSpec.swift
fdiaz: I don't think that's possible, the argument label is always present IIUC?

> Each function parameter has both an argument label and a parameter name. The argument label is used when calling the function; each argument is written in the function call with its argument label before it. The parameter name is used in the implementation of the function. 

I think you're referring to having the argument label be the same as the parameter name, which I think we already have tests for?

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: > Maybe what's confusing me more is the fact that we're setting the parameterName to be argumentLabelName if there is no explicit parameterName. In fact, doing this loses data.

I don't see a problem with scanning those 2 equally since they are, in fact, equal.

I think we've decided to be opinionated here in terms of style (e.g. using `Array` vs `[]`) so I think it's ok to lose the fact that the parameter name and argument labels are or are not explicitly defined.

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: I went back and re-read this.

Based on:

> Each function parameter has both an argument label and a parameter name. The argument label is used when calling the function; each argument is written in the function call with its argument label before it. The parameter name is used in the implementation of the function.

I'm again confused. I thought this meant that if we only defined one 👍🏼 
`func foo(bar: String)`

then bar is the argument label _and_ the parameter name. And then 
`func foo(bar bar2: String)`

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
dfed: they're equal from a "how would you call the code" POV, but not from a "how would you generate a protocol for this method" POV. Though I just tried it out and the following code does compile:
```
protocol TestProtocol {
    func doSomething(parameterName parameterName: String)
}

final class TestConformance: TestProtocol {
    func doSomething(parameterName: String) {
        // do something
    }
}
```

So I withdraw the objection, since this API is not limiting a consumer's ability to generate compiling code.

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: So I don't think 

> Technically speaking if a function argument only has a parameter name and no argument label then the single name is the parameter name

Is correct?

How I understand this is that:

- A _parameter_ has 2 members that _always exist_ an argument label and a name (i.e. the "parameter name")
- When we explicitly define only one "member" in the parameter, then the argument label is equal to the name.

So I think making either of these nil is just incorrect, since they can never technically be nil.

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: I just pushed a new commit that hopefully will help clarify a bit https://github.com/fdiaz/SwiftInspector/pull/136/commits/f4c285dd18f85cbd4fd352cb0026d2286ea6e77e

Sources/SwiftInspectorVisitors/Tests/FunctionDeclarationVisitorSpec.swift
fdiaz: I just pushed a new commit that hopefully will help clarify a bit https://github.com/fdiaz/SwiftInspector/pull/136/commits/f4c285dd18f85cbd4fd352cb0026d2286ea6e77e

Sources/SwiftInspectorVisitors/FunctionDeclarationVisitor.swift
fdiaz: Ok, this is what I ended up with https://github.com/fdiaz/SwiftInspector/pull/136/commits/f4c285dd18f85cbd4fd352cb0026d2286ea6e77e

fdiaz:
I'm gonna merge for now but do leave more feedback if you have it and we can address it later!

